PR #4, opened 2026-05-02
fix: preserve pin message duration when deleting messageContextInfo

zdanysfa:
# Fix Pin Message Support

## Description
Pin message functionality was broken because `messageAddOnDurationInSecs` (the pin duration field) was being deleted along with `messageContextInfo` before sending to the Rust bridge.

## Problem
In `Socket/messages.ts`, the code deleted the entire `messageContextInfo` object to avoid conflicts with Rust-generated reporting tokens and message secrets. However, `messageAddOnDurationInSecs` is required by WhatsApp to specify pin duration and is never set by Rust — so deleting it caused pin messages to fail silently.

## Root Cause
- `generateWAMessageContent()` in `Utils/messages.ts` correctly sets `messageAddOnDurationInSecs` inside `messageContextInfo` (lines 459-467)
- `Socket/messages.ts` then deletes the entire `messageContextInfo` (line 81) before encoding and sending to Rust
- This deletion was intentional — Rust fills reporting tokens and message secrets itself
- But it also wiped `messageAddOnDurationInSecs`, which Rust never sets, breaking pin messages

## Solution
Preserve `messageAddOnDurationInSecs` before deleting `messageContextInfo`, then restore it afterward — but only for `pinInChatMessage` content type. This allows Rust to safely manage its own security fields without losing the pin duration metadata.

## Changes
- **Modified**: `src/Socket/messages.ts` (lines 72-86)
  - Save `messageAddOnDurationInSecs` before delete if content type is `pinInChatMessage`
  - Restore it after delete to preserve pin duration
  
- **Added**: `src/__tests__/pin-message.test.ts` (7 test cases)
  - PIN_FOR_ALL (pin for all group members)
  - UNPIN_FOR_ALL (unpin for all)
  - Default pin duration (604800 seconds / 7 days)
  - Explicit pin duration
  - Key attachment validation
  - Timestamp generation
  - Message context info structure

## Testing
- ✅ All 55 tests pass (including 7 new pin message tests)
- ✅ TypeScript build passes (`npm run build`)
- ✅ No linting errors (`npm run lint`)

## Security & Safety
✅ **Verified Safe** — No security impact

- `messageAddOnDurationInSecs` is pure UX metadata (pin duration in seconds: 86400/604800/2592000/0)
- Not a cryptographic field, authentication token, or security-sensitive data
- Rust bridge does NOT read or set this field — it only handles reporting tokens and message secrets
- Fix is isolated to pin messages only — other message types unaffected

## Notes
This fix ensures pin messages work correctly while maintaining the security boundary where Rust handles encryption and reporting tokens.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserve pin duration metadata when sending pinned messages so durations (including zero) persist across the bridge.

* **Tests**
  * Added comprehensive tests for pin message generation and round-trip encoding/stripping behavior.
  * Extended test matcher API with numeric "less than" assertions (toBeLessThan, toBeLessThanOrEqual).

### DIFF
--- a/src/Socket/messages.ts
+++ b/src/Socket/messages.ts
@@ -69,8 +69,21 @@ export const makeMessageMethods = (ctx: SocketContext) => ({
 			}
 		}
 
-		// Rust handles messageContextInfo (reporting tokens, message secrets) internally
+		// Rust handles messageContextInfo (reporting tokens, message secrets) internally.
+		// Exception: pinInChatMessage relies on messageAddOnDurationInSecs to tell the
+		// server how long to pin (86400s / 604800s / 2592000s) or 0 to unpin. The bridge
+		// does not set this field, so we must preserve it across the delete.
+		const pinAddOnDuration =
+			contentType === 'pinInChatMessage'
+				? (msg as { messageContextInfo?: { messageAddOnDurationInSecs?: number } }).messageContextInfo
+						?.messageAddOnDurationInSecs
+				: undefined
 		delete (msg as Record<string, unknown>).messageContextInfo
+		if (pinAddOnDuration !== undefined) {
+			;(msg as Record<string, unknown>).messageContextInfo = {
+				messageAddOnDurationInSecs: pinAddOnDuration
+			}
+		}
 
 		let msgId: string
 		const msgBytes = encodeProto('Message', msg as Record<string, unknown>)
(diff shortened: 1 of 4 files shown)


zdanysfa:
I've tested this fix in my bot project and everything works perfectly. No issues found. @jlucaso1 please review.

jlucaso1:
@kkzaadev thanks for that. I'll review it soon

PR #5, opened 2026-05-04
feat: add Redis-backed BridgeStore implementation

Santosl2:


Salientekill:
And what if you don't have redis? What happens then?

Santosl2:
> And what if you don't have redis? What happens then?

It's just a sotre for those using redis. Anyone who doesnt want to use it can create their own store or use default store

PR #17, opened 2026-08-06
ci: automate releases with release-please and npm trusted publishing

jlucaso1:
Releases here are manual today: bump `package.json` by hand, then `npm publish` from a laptop. Two problems — the tarball depends on the state of that disk, and publishing needs a long-lived npm token.

## How it works

`release-please` keeps a release PR up to date from the conventional commits already used in this repo. Merging that PR tags the release, and **only that tag** triggers a publish — the version bump stays reviewable instead of firing on every merge to `main`.

The publish job authenticates to npm over **OIDC (trusted publishing)** rather than a stored token. Provenance attestations are attached automatically, so no `--provenance` flag and no `NODE_AUTH_TOKEN` anywhere.

## Choices worth flagging

- **release-please over semantic-release** — semantic-release publishes straight off the merge, with no reviewable bump. release-please matches the conventional commits already in use without asking for a changeset file on every PR.
- **Actions pinned to commit SHAs** — this workflow is the publish path, so a moving tag on a third-party action would be a supply-chain hole.
- **`npm-publish` environment** — already created on the repo. Add required reviewers there to gate releases; the same name scopes the OIDC trust on npm.
- **`npm install -g npm@latest`** — trusted publishing needs npm >= 11.5.1, and the version bundled with a Node release is just whatever shipped with it.
- **Suite re-runs against the tag** before publishing. Redundant with the PR checks, but publishing cannot be undone.

## Required before this can publish

One-time trusted publisher setup on npmjs.com for `@oxidezap/baileyrs`, naming this repository, `release.yml`, and the `npm-publish` environment. Until that exists the publish step fails — nothing else in CI is affected.

## Note on ordering

The manifest pins `0.0.33`, the version currently on `main`. #16 bumps to `0.0.34` by hand; if it merges first, the manifest needs to follow before this lands, otherwise release-please would propose a bump from the wrong base.

---
## Summary by cubic
Automates package releases with `release-please` and `npm` trusted publishing. Adds a safe, tag-based recovery path and blocks unmerged code from publishing.

- **New Features**
  - Auto-generated release PR via `release-please`; merging it creates the tag and triggers publish.
  - Optional GitHub App token to auto-run checks; otherwise `GITHUB_TOKEN` requires “Approve and run”.
  - Verify and publish both check out the exact commit `release-please` tagged; push runs refuse when the event SHA differs and instruct to dispatch on the tag.
  - Recovery via `workflow_dispatch` on a release tag only: commit must be reachable from `main`, a GitHub release must exist, the tag must match `package.json`’s version, and implicit `latest` is rejected if a higher version exists; a single global concurrency group avoids push/dispatch races.
  - Publish uses OIDC (no long‑lived token) with automatic provenance; Node 24 and `npm@11.19.0` pinned; `npm ci --ignore-scripts`, explicit build, and `scripts/check-pack.ts` validate package contents.
  - Only `feat`/`fix`/`perf` trigger a release; use `fix(deps): …` for dependency bumps; actions pinned to SHAs.

- **Migration**
  - One-time trusted publisher setup on `npm` for `@oxidezap/baileyrs`, linking this repo, `release.yml`, and the `npm-publish` environment.
  - `.release-please-manifest.json` is set to `0.0.34`; update if `main` differs.

<sup>Written for commit 837ef395c2da761c5a55f6cd4f9eac4d805c381c. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Added automated release creation when changes are pushed to the main branch.
  * Added package publishing with build, validation, and provenance checks.
  * Added automated version management and changelog generation.

* **Chores**
  * Configured release tags, versioning rules, and categorized changelog sections.


jlucaso1:
Applied the review findings, plus a rebase onto main now that #16 merged.

**Pinned npm** (`npm@11.19.0`, was `@latest`) — greptile P2 / codex P2. Valid: that binary runs `npm publish` with the OIDC token, so a floating tag puts unreviewed code on the privileged path. Pinned to the latest 11.x rather than 12.0.2 to stay on the major the runners already ship.

**`persist-credentials: false`** on checkout — the job runs third-party install and build scripts while holding publish rights, and nothing talks to the remote after checkout.

**Manifest bumped to `0.0.34`** to match main after #16.

---

**Not applied: the PAT for release PR checks** (codex P1). The mechanism is real and documented — GitHub suppresses workflow triggers for events made with `GITHUB_TOKEN`, so the release PR opens with no check runs. But:

- Neither consequence the finding describes currently applies: `main` has no branch protection, so the PR cannot be blocked waiting on required statuses, and there are no "asserted checks" being bypassed.
- The fix costs a long-lived PAT or GitHub App credential, which is exactly what the OIDC publish removes. Trading a scoped, short-lived token for a stored one to check a PR that only edits `version` and `CHANGELOG.md` is a bad trade.
- The publish job re-runs the full suite against the tag before publishing, so nothing ships unverified either way.

Documented in the workflow with the condition that would change the answer: if `main` ever gets required status checks, the release PR becomes unmergeable and a GitHub App token is then warranted.

jlucaso1:
Second round applied. Two of the three findings were already fixed in the previous push (`npm@latest` → `npm@11.19.0`, and the App-token path for release PR checks) — those comments predate it. The two that still stood:

## Publish credentials reachable by install/test code — applied

cubic P1 and greptile converged on this, and it was fair: `npm ci` runs dependency lifecycle hooks and the suite ran in the same job holding `id-token: write`, so any of it could mint a token and publish under this package identity.

Split into two jobs:

- **`verify`** (no `id-token`): `npm ci`, lint, format, tests.
- **`publish`** (`id-token: write`, needs `verify`): `npm ci --ignore-scripts`, explicit `npm run build`, `check-pack`, then `npm publish --ignore-scripts`.

Dependency install hooks and the test suite no longer run anywhere near the token. What is left in the privileged job is the build itself, which cannot be removed while the published content is generated.

I did **not** take the "hand a prebuilt tarball to a minimal publish job" route. It is the stronger version of the same idea, but npm documents provenance against publishing from the package directory and is silent on pre-packed tarballs — I was not willing to trade a confirmed provenance guarantee for an unverified one. Happy to revisit if that gets clarified.

Verified locally before pushing that `npm ci --ignore-scripts && npm run build && node scripts/check-pack.ts` works — skipping install hooks does not break the build here (`check-pack: 259 files, no duplicates`).

## `chore(deps):` would never release — applied

Good catch, and it would have bitten exactly the most common change in this repo. Only feat/fix/perf produce a release; this repo bumps the bridge as `chore(deps): …` (`fe9a641`, and my own `9c780b1`), which release-please types as `chore`. Such a bump would sit on main unpublished until an unrelated releasable commit dragged it out.

Also dropped the `deps` entry from `changelog-sections` — it was decorative, since no commit here is typed `deps`. Documented in the workflow header that dependency bumps go as `fix(deps): …`.

jlucaso1:
Third round. Two applied, two rejected with evidence.

## Applied

**`issues: write` on the release job** (coderabbit, Major). Correct — release-please labels the release PR, and PR labels go through the issues API. I had granted it on the App but forgotten it on the job `permissions`, which is what governs the `GITHUB_TOKEN` fallback path.

**Scope the App token with `permission-*`** (coderabbit, Major). Also correct, and a good catch about a non-obvious interaction: without those inputs the token inherits every permission the App installation holds, and job-level `permissions` do not constrain it. Today the installation is already minimal (contents / pull-requests / issues), so nothing changes right now — but it stops the release token from silently widening if the App is ever granted more for another purpose. Now pinned to those three.

## Rejected

**P1 "Pass the application identifier as app-id"** — this is wrong, and checkable at the exact SHA the finding cites. `action.yml` at `bcd2ba4`:

```yaml
inputs:
  client-id:
    description: "GitHub App Client ID"
    required: false
  app-id:
    description: "GitHub App ID"
    required: false
    deprecationMessage: "Use 'client-id' instead."
```

`client-id` is supported; `app-id` is the deprecated one. Keeping `client-id`.

**P2 "Move the bootstrap cutoff to the 0.0.34 commit"** — rests on two premises that do not hold here.

It says the first PR would "repeat already shipped changes". Those changes were never shipped: npm has `0.0.32` as the latest, while `main` is at `0.0.34`. The bumps in #15 and #16 exist only in git — no npm user has ever received them. That is precisely why the cutoff is `490bf47`, the last commit that actually corresponds to a published version.

It also says the bump "can jump 0.0.34 to 0.1.0". That needs a breaking change, and there is none in the interval:

```
$ git log 490bf47..main --format="%h %s%n%b" | grep -iE "BREAKING CHANGE|^[a-z]+(\(.+\))?!:"
(no output)
```

With `bump-patch-for-minor-pre-major`, `feat` + `fix` both map to patch, so the first release is `0.0.35` — carrying the two unpublished versions, with a changelog that finally documents them.

**"Remove executable package setup from the OIDC publisher job"** (coderabbit, and greptile's "build code retains publish authority") — same trade-off answered in the previous round: the alternative is publishing a prebuilt tarball, and npm documents provenance against publishing from the package directory while saying nothing about tarballs. Standing by the confirmed guarantee.

jlucaso1:
Valid, applied in `2855099`.

Two problems in one: `verify` and `publish` each resolved the tag name independently, so a tag moving in between would have them build different trees — and neither used the commit npm actually records. Provenance reads the event's sha regardless of what the workspace was checked out to, so a run could publish one tree while attesting another.

Both jobs now check out `${{ github.sha }}`, which cannot move. release-please tags the commit it just merged, so that sha and the tag agree — and `verify` now asserts it instead of assuming:

```yaml
- name: Confirm the release tag points at this commit
  env:
    GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
    TAG: ${{ needs.release-please.outputs.tag_name }}
  run: |
    ref=$(gh api "repos/$GITHUB_REPOSITORY/git/ref/tags/$TAG")
    target=$(jq -r .object.sha <<<"$ref")
    if [ "$(jq -r .object.type <<<"$ref")" = tag ]; then
      target=$(gh api "repos/$GITHUB_REPOSITORY/git/tags/$target" --jq .object.sha)
    fi
    ...
```

The annotated-tag dereference is not decorative: release-please creates annotated tags, so `/git/ref/tags` returns a tag object rather than the commit, and comparing that against `GITHUB_SHA` would fail every release. Tested both directions against the real `v0.6.1` tag on the sibling repo — resolves to the right commit, and exits 1 with a clear message when given a mismatched sha.

Also applied the same fix to the bridge's workflow (`oxidezap/whatsapp-rust-bridge#4`, commit `e4f66c7`), which had the identical pattern on its push path. Keeping one of the two with a hole I had just characterised did not seem defensible.

jlucaso1:
Both applied in `1125efb`.

**The event sha is not always the release commit** (P1). This one is right and I had it backwards. My comment claimed "release-please tags the commit it just merged, so the two agree" — true only when the run for that merge is the one that creates the release. When it fails or is superseded, a later push creates the release, the tag names the earlier merge, and my equality assertion fails *after* the GitHub release already exists. Worse than a false alarm: no later run reports `release_created`, so that version becomes unpublishable without hand intervention. I had turned a recoverable hiccup into a stuck release.

The action reports the commit it tagged (`steps.release.outputs.sha`, documented as "SHA that a GitHub release was tagged at"). Both checkouts and the assertion now use it, which also makes the check meaningful rather than circular: it verifies the action against itself instead of against a guess about which push arrived first.

**Retargeting during environment approval** (P2). Covered by the same change rather than by a recheck after the gate: both jobs pin a sha, so moving the tag while publish waits on its approval cannot change what gets published. A recheck would only tell you the tag no longer matches — the package would already be correct. Enforcing immutable release tags is worth doing at the repo level, but it is a ruleset, not a workflow change.

Same fix applied to the bridge (`oxidezap/whatsapp-rust-bridge#4`, `adf85df`), which had the identical assumption.

PR #26, opened 2026-08-08
fix(deps): update whatsapp-rust-bridge to 0.7.0

jlucaso1:
## Summary

`@oxidezap/whatsapp-rust-bridge` 0.6.5 to 0.7.0.

Nothing new is exposed. 0.7.0 adds roughly fifty operations (newsletter, labels, quick replies, business catalog, server queries, privacy and tokens); none of them is wired up here. This PR is only the bump plus the four runtime behavior changes that came with it.

Observable for a consumer of this package:

- A mistyped argument to a bridge-backed socket method now rejects with `invalid-argument` naming the parameter, where it used to leave the promise pending forever and end the Node process with an uncaught exception.
- An app-state mutation rejected for the caller's own input (mute, pin, archive, star, contact names) now reports `invalid-argument` instead of `internal`.
- `logged_out.reason` is now a written-down string rather than a `Debug` rendering. The values did not change, so no consumer comparison breaks.
- `encryptMediaStream` rejects a non-stream instead of accepting it and failing much deeper.

Type surface: I diffed the emitted `.d.ts` between the two versions with comments stripped. There is no removal, no signature change and no shape change to any exported type. The only deletions are wasm-bindgen glue symbols in `InitOutput`. `npm run lint` (which runs `tsc`) passed on the first try across the roughly 60 import sites, and the `WhatsAppEvent` union is byte-identical, so `Bridge/__tests__/dts-drift.test.ts` needed nothing.

### Public strings no longer come from `Debug` (bridge #28)

`logged_out.reason`, newsletter `verification`/`state`, and newsletter follower `role` used to cross as `format!("{:?}", ...)`. `Debug` is not a stable API, so a variant renamed in the core would have silently changed a string consumers compare, with nothing failing to build on either side.

The important detail, and the one the pre-research guessed the other way: **this is a change of source, not of value.** The mappings emit exactly what `Debug` emitted (`"LoggedOut"`, `"TempBanned"`, `"Verified"`, `"Active"`, `"Owner"`, and `Unknown(499)`-style values for the core's own fallback). Nothing arriving here changed.

What this repository does with it: `evt.reason` is consumed in one place, `src/Socket/events.ts:392`, where it is interpolated into the close message. It decides nothing; the status code is always `DisconnectReason.loggedOut`. `connect_failure.reason` is a separate field and a `number`, so `isReconnectableConnectFailure` is untouched.

Because the string is now a contract rather than an accident, `src/__tests__/auto-reconnect-terminal-close.test.ts` pins all fifteen spellings against the close message a consumer sees, plus the no-reason case. The events are synthetic, so to be precise about what that buys: it pins our half of the contract, that the dispatcher passes the reason through untouched, that the message is exactly the literal `Logged out: ` followed by the reason, and that the status code stays `loggedOut` whatever the reason says. A change on the bridge side of the line is caught by the bridge's own tests, which assert those spellings literally. This test therefore passes on 0.6.5 too, since the values are the same on both.

### A mistyped parameter rejects instead of escaping the caller (bridge #25)

Twenty-four typed parameters were deserialized by tsify's own `FromWasmAbi`, inside the async shim, where a throw escapes as an uncaught exception. The result was the worst of both: the promise never settled, and the process died. 0.7.0 takes them as `JsValue` and deserializes explicitly, rejecting with `invalid-argument` and the parameter name.

The accepted set of values is unchanged (same serde deserialization, different call site), so no valid call started failing.

Our own call sites are type-correct and I checked each one that touches an affected parameter: `readMessages`/`markPlayed` go through `receiptMessageKeys`, which already projects to the exact `{remoteJid, id, participant?}` shape and drops keys missing either required field; `addLidPnMappings` filters before calling; `muteChat`, `sendPresence`, `sendChatState`, the group and block actions, and the media-type parameters are all pass-throughs of typed values. Nothing here needed changing.

`src/__tests__/bridge-argument-rejection.test.ts` covers the new contract: four probes against a real offline bridge client, in a child process so a rejection can be told apart from a call that ends the runner. All of them fail on 0.6.5 and pass on 0.7.0.

They drive the bridge boundary directly rather than the socket methods layered over it, which matters for reading them honestly: `sock.readMessages` filters its input through `receiptMessageKeys` first, so that particular bad value resolves there without ever reaching the bridge. The boundary is still consumer-facing, since `MediaGenerationOptions.processMedia` hands the raw client over.

### A bad argument reports `invalid-argument`, not `internal` (bridge #29)

Six rejections that were about the caller's input said `internal`. The broad one is `AppStateError::InvalidRequest`, which covers every app-state mutation: mute, pin, archive, star, labels, contact names.

Nothing in `src/` branches on a bridge error `kind`; I grepped for `kind`, `serverCode` and code comparisons, including the tests. `src/Utils/generics.ts`'s `getCodeFromWSError` reads transport errors, not bridge errors. So the impact here is entirely in what an end user sees, and no code changed.

Two narrower cases in the same PR do change what a call does. `muteChat` now range-checks its timestamp before the `as i64` cast, so `Infinity` rejects with `invalid-argument` on `muteUntil` instead of saturating to `i64::MAX` and passing every check downstream. That one is covered in the rejection test above.

### `encryptMediaStream` rejects a non-stream (bridge #30)

`input` and `output` were imported class types, which wasm-bindgen casts unchecked, so a plain object was accepted at the boundary and blew up deep inside the stream machinery as another uncaught exception with the call pending forever. 0.7.0 probes each one by calling `getReader()`/`getWriter()` and releasing the lock, and rejects naming the parameter.

This repository never calls `encryptMediaStream` itself. It reaches consumers through `MediaGenerationOptions.processMedia`, which is handed `waClient` typed as a `Pick` of `WasmWhatsAppClient` over `uploadMedia`, `encryptMediaStream` and `uploadEncryptedMediaStream` (`src/Types/Message.ts:359`). The rejection test covers the parameter contract; the happy path with real web streams lives in `src/__tests__/e2e/send-receive-message.test-e2e.ts`, which I could not run locally but which CI ran green on this branch.

Note for anyone passing streams through `processMedia`: the probe consumes and releases a reader/writer, so an already-locked stream is now rejected rather than accepted.

## Bugs this surfaced

One test failure, and it was not a bug in this repository.

`src/__tests__/bridge-free-safety.test.ts` asserted that `free()` with a call in flight dies with dlmalloc's `assertion failed: psize <= size + max_overhead`. On 0.7.0 it dies with `RuntimeError: memory access out of bounds` instead.

I checked before touching it. The assertion string is still compiled into both wasm builds, and both crashes are deterministic across repeated runs (5/5 each). The hazard is unchanged: `free()` with a pending call still corrupts the heap, still ends the process, and awaiting `disconnect()` first still makes it safe, which the suite's other two cases confirm on both versions. What changed is which memory check trips first, a property of the larger 0.7.0 heap layout, not of the invariant `end()` depends on.

So the test was pinning a build detail. It now accepts either fault and additionally requires the stack to come from inside wasm, which the old assertion did not check. That second condition is new coverage, not slack: a crash raised in JS would previously not have been distinguished.

No assertion was loosened elsewhere, no timeout raised, nothing skipped. The other 645 tests passed unchanged.

## Reviewed and no change needed

- **`src/Types/Newsletter.ts` unions.** The pre-research suspected `verification?: 'VERIFIED' | 'UNVERIFIED'` and `mute_state?: 'ON' | 'OFF'` never matched what the bridge sends, and that 0.7.0 might fix a latent bug. Neither is the case. `NewsletterMetadata` in that file is a mirror of upstream Baileys' type, kept for API compatibility, and is not what `sock.newsletterMetadata()` returns: that returns the bridge's `NewsletterMetadataResult`, where `verification` and `state` are plain `string`. Nothing in `src/` compares them to those literals. The unions are correct as upstream mirrors and wrong to "fix" here; the real gap is that `newsletterMetadata()` does not project onto them, which belongs to the newsletter domain, not to a version bump.
- **`connect_failure.reason`.** A `number` (`ConnectFailureReason` wire code), untouched by #28. `isReconnectableConnectFailure` and `mapConnectFailureToDisconnect` in `src/Socket/events.ts` keep working on the same values.
- **Error-code handling.** No path in `src/` reads a bridge error's `kind`, so #29 cannot change control flow here. The one test that asserts on a bridge error (`src/__tests__/group-operations-compatibility.test.ts`) checks `serverCode`, which #29 does not touch.
- **`getMediaConn` doc drift.** The 0.6.5 doc comment claimed `hosts` carried `maxContentLengthBytes`; 0.7.0 corrected the comment. `MediaHost` itself is `{ hostname: string }` in both, so nothing here was relying on a field that was never there.
- **Module split (bridge #26).** `wasm_client.rs` was split per domain. Every type still resolves from the same import path: `dist/whatsapp_rust_bridge.d.ts` is the only declaration file that changed at all, and the comment-stripped diff shows no declaration moved, was removed, or changed shape.
- **`src/Socket/index.ts` logout path.** Builds its own `'Logged out'` Boom independently of `evt.reason`, so the dispatcher's message and this one stay separate as before.

## Coverage

`npm run compat:audit` after: **97.15% (21717/22354)**

Identical across all four categories (exports 85.31%, functions 96.04%, fields 98.00%, events 100.00%). It should be: the audit measures this package's public surface against upstream Baileys, and this PR adds no public method. A move here would have meant a new surface slipped in.

## Follow-ups

Deliberately unused, one domain per PR:

- **Newsletter** (`compat-newsletter.md`): the remaining newsletter operations from bridge #23, plus projecting `newsletterMetadata()` onto the upstream `NewsletterMetadata` shape noted above.
- **Labels, contacts, quick replies** (`compat-labels-contacts-quickreply.md`): `addChatLabel`, `addMessageLabel` and the rest of bridge #21's app-state label actions. The new `QuickReplyUpdate` type arrived with them but no matching `WhatsAppEvent` variant, worth checking how it is delivered before building on it.
- **Business catalog** (`compat-business-catalog.md`): bridge #24, including `getOrder` and the new `ProductAvailability` type.
- **Server queries** (`compat-server-queries.md`): bridge #20's lifecycle and server-side operations, e.g. `cleanDirtyBits`.
- **Privacy and tokens** (`compat-privacy-and-tokens.md`) and **socket internals** (`compat-socket-internals.md`).

Also worth knowing when those land: `assertSessions`' `force` and `getUSyncDevices`' `useCache`/`ignoreZeroDevices` are documented in 0.7.0 as having no effect. The parameters existed before; only the documentation is new.

## Release type

The bump is committed as `fix(deps)`, not `chore(deps)`, and the PR title matches. `.github/workflows/release.yml` records the reason: only feat/fix/perf produce a release, and a chore-typed bridge bump would leave the upgrade sitting on main until an unrelated commit dragged it out.

## Validation

Run locally and green:

- `npm run format`
- `npm test`: 666 tests across 167 suites, 0 failures. 645 across 165 before this PR, so 21 added and none removed.
- `npm run compat:audit`

`npm run test:e2e` could not run locally, since it needs a mock server this environment does not have. CI does have it, and the E2E workflow passed on this branch, which is what covers the `encryptMediaStream` happy path with real `ReadableStream`/`WritableStream` values. All four workflows (Build, Run Tests, Linting, E2E) are green.

Cross-version checks done by hand, against real bridge builds:

- The four rejection probes: 5 failures on 0.6.5, 5 passes on 0.7.0.
- The free-safety crash signature: 3 runs on 0.6.5 all dlmalloc, 5 runs on 0.7.0 all out-of-bounds, and the reworked test passes on both.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@hapi/boom": "^9.1.4",
-        "@oxidezap/whatsapp-rust-bridge": "0.6.5",
+        "@oxidezap/whatsapp-rust-bridge": "0.7.0",
         "long": "^5.3.2",
         "pino": "^10.3.1",
         "protobufjs": "^7.6.5"
@@ -974,9 +974,9 @@
       }
     },
     "node_modules/@oxidezap/whatsapp-rust-bridge": {
-      "version": "0.6.5",
-      "resolved": "https://registry.npmjs.org/@oxidezap/whatsapp-rust-bridge/-/whatsapp-rust-bridge-0.6.5.tgz",
-      "integrity": "sha512-vRkO2vWAG79DWGxp5ffoUvBENO6WloTh1+u1Tsf/s+LQ/VmUAIsCSAuk996MsnmFbCrtKauhSx82EPGWXlt88Q==",
+      "version": "0.7.0",
+      "resolved": "https://registry.npmjs.org/@oxidezap/whatsapp-rust-bridge/-/whatsapp-rust-bridge-0.7.0.tgz",
+      "integrity": "sha512-RD8Ueg+xf2MAIUoYz+rlkHvWyOOEqvafEH6nh7xARq6JvGOcWBYy9xxDwRy03HWP9lVeIYUPeY5lqRYaTH9W6g==",
       "license": "MIT"
     },
     "node_modules/@oxlint/binding-android-arm-eabi": {
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
   },
   "dependencies": {
     "@hapi/boom": "^9.1.4",
-    "@oxidezap/whatsapp-rust-bridge": "0.6.5",
+    "@oxidezap/whatsapp-rust-bridge": "0.7.0",
     "long": "^5.3.2",
     "pino": "^10.3.1",
     "protobufjs": "^7.6.5"
--- a/src/__tests__/auto-reconnect-terminal-close.test.ts
+++ b/src/__tests__/auto-reconnect-terminal-close.test.ts
@@ -234,11 +234,12 @@ describe('connection.update: terminal disconnects', () => {
  * with nothing here failing to build. Bridge 0.7.0 writes the spellings down
  * instead, keeping every value it already emitted.
  *
- * That makes them a contract, and this is where the contract is observed: the
- * reason is interpolated straight into the close message, so it is the text a
- * consumer logs and greps. It steers nothing, the status code is always
- * `loggedOut`, but a change to it is a change to the public surface, and
- * without this test the next one would pass unnoticed too.
+ * The events are synthetic, so this pins our half of that contract rather than
+ * the bridge's: the dispatcher hands the reason through untouched, the close
+ * message is exactly `Logged out: <reason>`, and the status code stays
+ * `loggedOut` whatever the reason says. The list doubles as the record of which
+ * values the bridge undertakes to emit. A change on the bridge side of the line
+ * is caught by the bridge's own tests, which assert those spellings literally.
  */
 const LOGGED_OUT_REASONS = [
 	'Generic',
--- a/src/__tests__/auto-reconnect-terminal-close.test.ts
+++ b/src/__tests__/auto-reconnect-terminal-close.test.ts
@@ -227,6 +227,56 @@ describe('connection.update: terminal disconnects', () => {
 	}
 })
 
+/**
+ * `logged_out.reason` used to be `format!("{:?}")` on the core's
+ * `ConnectFailureReason`, a `Debug` string, which is not a stable API: a
+ * variant renamed upstream would have silently changed what a consumer reads,
+ * with nothing here failing to build. Bridge 0.7.0 writes the spellings down
+ * instead, keeping every value it already emitted.
+ *
+ * That makes them a contract, and this is where the contract is observed: the
+ * reason is interpolated straight into the close message, so it is the text a
+ * consumer logs and greps. It steers nothing, the status code is always
+ * `loggedOut`, but a change to it is a change to the public surface, and
+ * without this test the next one would pass unnoticed too.
+ */
+const LOGGED_OUT_REASONS = [
+	'Generic',
+	'LoggedOut',
+	'TempBanned',
+	'AccountLocked',
+	'UnknownLogout',
+	'ClientOutdated',
+	'BadUserAgent',
+	'CatExpired',
+	'CatInvalid',
+	'NotFound',
+	'ClientUnknown',
+	'InternalServerError',
+	'Experimental',
+	'ServiceUnavailable',
+	// The core's own fallback for a code it cannot name, carried through.
+	'Unknown(499)'
+]
+
+describe('logged out: the reason reaches the consumer verbatim', () => {
+	for (const reason of LOGGED_OUT_REASONS) {
+		it(`${reason} → 'Logged out: ${reason}'`, () => {
+			const result = dispatch({ type: 'logged_out', data: { reason } })
+
+			expect((result.close?.lastDisconnect?.error as Boom | undefined)?.message).toBe(`Logged out: ${reason}`)
+			expect(result.statusCode).toBe(DisconnectReason.loggedOut)
+		})
+	}
+
+	it('no reason at all → plain "Logged out"', () => {
+		const result = dispatch({ type: 'logged_out', data: {} })
+
+		expect((result.close?.lastDisconnect?.error as Boom | undefined)?.message).toBe('Logged out')
+		expect(result.statusCode).toBe(DisconnectReason.loggedOut)
+	})
+})
+
 describe('connection.update: disconnects the engine retries', () => {
 	for (const { label, event } of RETRYING) {
 		it(`${label} → 'connecting', never 'close'`, () => {
--- a/src/__tests__/bridge-argument-rejection.test.ts
+++ b/src/__tests__/bridge-argument-rejection.test.ts
@@ -8,12 +8,22 @@
  * rejects with `invalid-argument` naming the parameter, and it does the same
  * for a stream parameter that cannot do what a stream does.
  *
- * What this pins is the bridge boundary itself, driven directly, not the socket
- * methods layered over it. The distinction is real: `sock.readMessages` runs its
- * input through `receiptMessageKeys` first, which drops anything without a
- * `remoteJid` and an `id`, so this exact bad value resolves there without ever
- * reaching the bridge. The boundary is still a consumer-facing contract, since
- * `MediaGenerationOptions.processMedia` hands the raw client over.
+ * What this pins is the bridge boundary itself, driven directly. How much of
+ * the socket surface each probe stands for varies, so to be exact about it:
+ *
+ *  - `muteChat` is a bare pass-through on the socket, so that probe is the
+ *    public path, argument and all.
+ *  - `encryptMediaStream` reaches consumers raw, handed over by
+ *    `MediaGenerationOptions.processMedia`.
+ *  - `readMessages` is filtered by `receiptMessageKeys` first, which drops
+ *    anything without a `remoteJid` and an `id`, so through the socket this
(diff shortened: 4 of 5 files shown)


jlucaso1:
Retyped the bump from `chore(deps)` to `fix(deps)`, title included. The Codex P1 was right and it is the one finding here with a shipping consequence: `release.yml` says in so many words that a chore-typed bridge bump releases nothing and leaves the upgrade sitting on main until an unrelated commit drags it out. Same tree, only the messages changed, so the force-push carries no content.

While retyping I also moved `fix(tests): describe the free() hazard...` to `test(bridge): ...`. It changes no shipped code, and `fix` would have listed it in the user-facing changelog as a bug fix.

Everything else that landed since the first push:

- A stalled probe now exits the child immediately. It used to leave its call in flight and carry on to `free()`, which is the heap corruption the sibling suite documents, so the crash got reported instead of the probe that caused it.
- Added a `child.on('error')` handler. A failed spawn was throwing in the runner and leaving the promise pending until the suite timeout.
- Both assertions now carry the child's stderr. It was being collected and never shown, so a child that died during setup failed with `expected false to be true`.
- Reports are written with an awaited write callback instead of `console.log`. stdout is a pipe, and `process.exit()` does not wait for it, so a loaded runner could truncate the run and the parent would read that as a regression.
- Dropped `unreachable` from the fault allowlist. The other two entries name the build that emits them; that one named none.

Two comments I did not act on, with reasons.

**Deriving the logged-out reasons from the bridge instead of injecting them.** Fair as stated, and I have corrected the comment rather than the test: a real `logged_out` event needs a server saying so, which is e2e territory. What the test can pin is our half, and that is now what it claims to pin: the dispatcher passes the reason through untouched, the message is exactly `Logged out: <reason>`, and the status code stays `loggedOut` regardless. The bridge asserts the spellings literally in its own tests.

**Running the rejection probes through `makeWASocket`.** The reading of `receiptMessageKeys` is correct: `sock.readMessages('not-an-array')` iterates the characters, drops every one for lacking `remoteJid` and `id`, and resolves without touching the bridge. So going through the socket would test our normalizer, not the bridge change that is the subject here. I have reworded the comment to say the probes drive the boundary directly and to name that filtering explicitly, and to note the boundary is still consumer-facing because `MediaGenerationOptions.processMedia` hands the raw client over.

**Setting an explicit `cwd` on the spawned child.** Skipped. Node resolves the bare specifiers by walking up from the eval base, so any cwd inside the repo finds `node_modules`, and a cwd outside it means `node --test` matched no test files to begin with. Deriving the root as `SRC/../..` would add an assumption about where the file sits that the code does not currently make.

`npm run lint` and `npm test` (666 tests) green on the retyped branch.

---

jlucaso1:
Checked all three against the branch as it stands.

**`chore(deps)` on the bump — already fixed, before this review landed.** The commit is `d6034c1 fix(deps): update whatsapp-rust-bridge to 0.7.0` and the PR title matches. The finding reads the file diff, which carries no commit type, so it cannot see this either way. Worth saying that the underlying point was right and it was Codex that caught it first; it is the one issue in this PR that would have cost a release.

**The logout-reason comment — already rewritten in 249cbe0.** The quoted phrases ("where the contract is observed", the claim that a bridge-side change would be caught) are gone. What is there now says the events are synthetic and that this pins our half only: the dispatcher passes the reason through untouched, the message format, and the status code staying `loggedOut`. Which is the same correction the finding asks for.

**The probes — half right, and I have made the comment exact rather than restructured the test.** Checked each of the four against its socket wrapper:

- `muteChat` is a bare pass-through (`chat-actions.ts:9`), so `sock.muteChat(jid, Infinity)` reaches the bridge with the bad argument intact. That probe *is* the public path.
- `encryptMediaStream` is handed to consumers raw through `MediaGenerationOptions.processMedia`.
- `readMessages` is absorbed by `receiptMessageKeys`, exactly as described.
- `sendPresence` is absorbed by `sendPresenceUpdate`, though not by filtering: `'nope'` is neither `available` nor `unavailable`, so it falls through to the branch that throws a missing-target-jid Boom (`index.ts:890`) before the bridge is reached.

So the blanket claim does not hold for half the probes, and the previous comment undersold them by lumping all four together. `3d4b4c6` names each one. I am not routing the absorbed two through `makeWASocket`: that would test our normalizers, not the bridge change this file is about, and for `readMessages` it would assert a silent no-op rather than a rejection.

Assertions unchanged, `npm run lint` and `npm test` green.

---

PR #28, opened 2026-08-08
feat(privacy): add call and message privacy controls

jlucaso1:
## Summary

Two methods, one line each. `updateCallPrivacy` sends category `calladd`, `updateMessagesPrivacy` sends `messages`, both through the generic privacy IQ this package already had. Nothing was needed in the core or the bridge.

A third method, `issuePrivacyTokens`, is declared as a no-op that warns once. Actually issuing a token would be a regression.

The privacy family moved out of the socket builder into `src/Socket/privacy.ts` first. That is not cosmetic: inline in `makeWASocket` is the one place in this package a method cannot be reached from a test, and the entire contract of these wrappers is which category string comes out the other side.

## Layer classification

| item | class | why |
|---|---|---|
| `updateCallPrivacy` | **(a)** | Core enumerates `PrivacyCategory::CallAdd` with wire value `calladd`; the bridge passes category and value straight through. |
| `updateMessagesPrivacy` | **(a)** | Same, `PrivacyCategory::Messages` with wire value `messages`. |
| `updateDisableLinkPreviewsPrivacy` | **(a), deferred** | Not a privacy IQ despite the name. It is an app-state mutation under index `['setting_disableLinkPreviews']`, and it lands in the silent `else` branch of `chatModify` today. It belongs to the labels task mechanically, and that is where I am doing it, so the two do not both touch that branch. |
| `issuePrivacyTokens` | **declared, no-op** | See below. |

Verified against `wacore/src/iq/privacy.rs` rather than assumed: `PrivacyCategory` carries `#[wire = "calladd"]` and `#[wire = "messages"]` alongside the six categories already wrapped here.

## Types

`WAPrivacyCallValue` and `WAPrivacyMessagesValue` already existed in `src/Types/Chat.ts`, and already match upstream's declarations character for character (`'all' | 'known'` and `'all' | 'contacts'`). No type work was needed.

One divergence worth recording, in the safe direction. The core documents `calladd` as accepting `all | known | contacts`, three values, while upstream's `WAPrivacyCallValue` has two. So our type is narrower than the server accepts. That costs a caller the ability to express `contacts` through the typed wrapper, and they can still reach it through `updatePrivacySetting('calladd', 'contacts')`. The failure mode this prompt warned about is the opposite one, a local type wider than the server accepts, which turns a compile error into a runtime error. That is not the case here.

## Deferred

**`updateDisableLinkPreviewsPrivacy`** goes to `fix/chat-modify-silent-noop`, for the reason in the table above. Recording the coordination explicitly so the two PRs do not collide on the same `chatModify` branch.

**`issuePrivacyTokens` does nothing, deliberately, and no gap prompt is being opened for it.** I checked the claim rather than taking it: `maybe_include_tc_token` is called from the send path (`src/send/mod.rs:2420`) on every 1:1 send, and `should_issue_tc_token` rate limits issuance through a sender bucket. The bridge exposes no token-issuing method, which is consistent with that being deliberate. A manual call would issue a second token outside the bucket and move the timestamp the limiter reads.

It first rejected, and a review pointed out that was wrong. The rule I was missing, which I will apply consistently across the rest of this series:

- **Throw when the caller's intent is not achieved.** A `chatModify({addChatLabel})` that resolves while nothing syncs is silent data loss and has to fail loudly. Making it do so is the point of the next PR.
- **Do nothing when the intent is already achieved by other means.** By the time a caller asks for privacy tokens, the engine has issued them. Rejecting aborts an upstream send workflow that awaits this mid-flight, breaking a call that was going to succeed, and buys nothing.

It warns once per socket rather than per call, so someone migrating from upstream finds out they can delete the call without a send loop flooding their log.

## Coverage

after: **97.22% (21732/22354)**

+15 matched declarations from three methods, because each one lands in several of the aggregate socket types the auditor walks.

One honest note on the shape of that delta. Missing dropped (functions 243 → 234, fields 235 → 226) while mismatches rose slightly. The new mismatches are not new incompatibilities: adding any method to the socket changes the giant inline return type of `makeWASocket`, and the auditor compares that as a string, so it re-reports it wherever it appears. The three methods themselves count as clean matches, which I confirmed by grepping the detailed report for them: zero hits under either `missing` or `mismatch`.

## Validation

- `npm run format`
- `npm test`: 679 tests, 0 failures, up from 666 with no existing test adapted
- `npm run compat:audit`

Not run: `npm run test:e2e`, which needs a mock server this environment does not have. Nothing in this PR is e2e-observable.

Each new assertion was mutation checked rather than assumed to bite: changing `calladd` to `call` fails the suite, and making `issuePrivacyTokens` stop warning fails it too. Both were verified by making the edit and running the tests, not by reading the assertion.

### DIFF
--- a/src/Socket/index.ts
+++ b/src/Socket/index.ts
@@ -27,11 +27,7 @@ import type {
 	ReachoutTimelockState,
 	SignalKeyStoreWithTransaction,
 	UserFacingSocketConfig,
-	WAPrivacyGroupAddValue,
-	WAPrivacyOnlineValue,
-	WAPrivacyValue,
 	WABusinessProfile,
-	WAReadReceiptsValue,
 	WAMessage,
 	WAMessageKey
 } from '../Types/index.ts'
@@ -57,6 +53,7 @@ import { makeMessageMethods } from './messages.ts'
 import { makeNewsletterMethods } from './newsletter.ts'
 import { makePreKeyMethods } from './prekeys.ts'
 import { makePresenceMethods } from './presence.ts'
+import { makePrivacyMethods } from './privacy.ts'
 import { makeProfileMethods } from './profile.ts'
 import { mapReachoutTimelock } from './reachout.ts'
 import { makeHttpClient, makeTransport } from './transport.ts'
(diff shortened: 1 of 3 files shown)


jlucaso1:
Good catch, and I have changed it. `issuePrivacyTokens` now resolves instead of rejecting.

The distinction I had missed is worth stating, because the same question comes up for every unsupported method in this compatibility series and I was about to apply the opposite rule in the next PR:

- **Throw when the caller's intent is not achieved.** A `chatModify({addChatLabel})` that resolves while nothing syncs is silent data loss. That one has to fail loudly, and making it do so is the whole point of the next PR.
- **Do nothing when the intent is already achieved by other means.** The engine issues privacy tokens on every 1:1 send, rate limited by a sender bucket. By the time a caller asks, the thing they wanted has already happened. Rejecting aborts a send workflow that awaits this mid-flight, which turns a compatible call into a broken one, and it does that to buy nothing.

I had lumped both under "declare and fail" because the task prompt suggested it, without noticing they are not the same shape of problem.

It still warns, once per socket rather than per call, so someone migrating from upstream finds out they can delete the call and a send loop does not flood their log. Test updated to match: it asserts the call resolves, that nothing reaches the bridge, and that three calls produce exactly one warning.

Body updated. `npm test` 679 passing, audit 97.22%.

---

PR #29, opened 2026-08-08
fix(chats): stop silently dropping unsupported chatModify variants

jlucaso1:
## Summary

`chatModify` ended in an `else` that logged a warning and resolved, for every variant the bridge did not expose: link previews, the five label actions, quick replies. `contact: null` did the same in its own branch.

Two changes, one commit each.

**The silent success is gone.** Every one of those variants now reaches the bridge call that actually performs it, so the `else` is only reachable by a modification this package does not know at all, and that rejects naming the variant.

**The ten missing public methods are here**, with upstream's signatures, each routed through `chatModify` exactly as upstream routes them: `addLabel`, `addChatLabel`, `removeChatLabel`, `addMessageLabel`, `removeMessageLabel`, `addOrEditContact`, `removeContact`, `addOrEditQuickReply`, `removeQuickReply`, `star`.

`updateDisableLinkPreviewsPrivacy` is deliberately **not** here even though it is in the privacy task's list. It is an app-state mutation, not a privacy IQ, so its `disableLinkPreviews` branch belongs to this PR. #28 records the same split from the other side so the two do not both touch that branch.

## Behavior change

A call that used to resolve now throws, and that is the point.

Before: `sock.chatModify({ addChatLabel: { labelId } }, jid)` resolved successfully, logged a warning, and synced nothing. A caller reasonably believed the chat was labelled.

After: it labels the chat. And a variant with no path at all rejects with `chatModify: unsupported modification '<name>'` rather than pretending.

Who this breaks: anyone whose code path depends, without knowing it, on those calls not throwing. Concretely, code that labels a chat inside a `try` that swallows errors will now take the catch branch. That is strictly better information than the silence it replaces, but it is a real behavior change and pre-1.0 is the moment to make it.

`contact: null` is the one worth calling out separately, because "works approximately" would have been worse than failing. It now goes to `removeContact`, not to `saveContact` with empty fields. Removal is the only contact mutation the wire models as a `Remove`; a `Set` carrying empty values is applied as a rename to the empty string. Getting that wrong corrupts the user's contact instead of deleting it.

I did not add a `CHANGELOG.md` note: the changelog here is generated by release-please from commit subjects, and the `fix(chats):` commit carries this.

## Layer classification

Everything is **(a)**, implementable here by delegation. Nothing needed the core or the bridge, and no gap prompt was opened.

| upstream variant / method | bridge call | note |
|---|---|---|
| `disableLinkPreviews` | `setLinkPreviewsDisabled(bool)` | |
| `addLabel` | `createLabel(id, name, color)` / `deleteLabel(id)` | one upstream variant, two wire actions: an edit carrying `deleted` is the delete |
| `addChatLabel` / `removeChatLabel` | same names | **arguments swap**: upstream `(jid, labelId)`, bridge `(labelId, chatJid)` |
| `addMessageLabel` / `removeMessageLabel` | same names | **arguments reorder**: upstream `(jid, messageId, labelId)`, bridge `(labelId, chatJid, messageId)` |
| `quickReply` | `setQuickReply(...)` / `deleteQuickReply(id)` | |
| `contact` (value) | `saveContact(jid, fullName, firstName, saveOnPrimaryAddressbook)` | |
| `contact: null` | `removeContact(jid)` | separate method by protocol necessity, see above |
| `star` (list) | `starMessage`, once per message | |

The two argument-order rows are the ones I would expect a reviewer to check hardest. Both sides are strings, so a swap type checks cleanly and silently labels the wrong thing. There is a test per row pinning the order.

**`star` in batch stays our work.** The wire has no batch form: the real client emits one mutation per message, and the core matches that. Iterating `starMessage` is the correct behavior, not a workaround.

## The quick reply identity question

The prompt flagged this as decisive, so I checked it rather than mapping by name. Upstream identifies a quick reply by `timestamp`; the bridge takes `id`. They are the same value.

Upstream builds the app-state index as `['quick_reply', mod.quickReply.timestamp || String(Math.floor(Date.now() / 1000))]`, so its `timestamp` is the index key, defaulted to a unix second when absent rather than being a time in its own right. The core takes the same string and passes it in the same position: `send_app_state_action(&schemas::QUICK_REPLY, &[id], &value)`.

Deleting is confirmed to be the same upsert with `deleted = true` and the payload cleared, not a `Remove`, which is why `removeQuickReply` maps onto the `quickReply` variant rather than a delete action of its own.

## Deferred

Nothing. Every item in this task had a path, so no gap prompt was written for either the bridge or the core.

Worth recording for whoever picks up the neighbouring tasks: the bridge exposes app-state actions **typed per action and no generic**. That is a deliberate boundary choice, and it is why upstream's `appPatch` still has no path. It belongs to the socket-internals task, and the answer there is likely to be a refusal.

## Coverage

after: **97.41% (21774/22354)**

+57 matched declarations. Missing dropped from 243 to 213 (functions) and 235 to 205 (fields).

The `chatModify` behavior change contributes nothing to that number, which is worth stating plainly: the signature was already correct, only the behavior was wrong, and no coverage metric was ever going to catch it. The audit moved because of the ten methods.

## Validation

- `npm run format`
- `npm test`: 691 tests, 0 failures, up from 666
- `npm run compat:audit`

Not run: `npm run test:e2e`, which needs a mock server this environment does not have.

**No existing test needed adapting, and that is itself a finding.** The prompt expected the current `chatModify` tests to need changing. There were none: `grep` for `chatModify` across the test suite returns only the file this PR adds. A branch that resolved without doing anything survived precisely because nothing exercised it.

Every new assertion was mutation checked by making the edit and running the suite, not by reading the assertion:

- restoring the silent `else` fails 2 tests
- swapping `addChatLabel` back to upstream's argument order fails 2 tests
- routing `contact: null` through `saveContact` with empty fields fails 2 tests

---
_Generated by [Claude Code](https://claude.ai/code/session_019n5h4gDvrNFW4nCnf2YpMw)_

---
## Summary by cubic
Stops silently ignoring unsupported `chatModify` variants by wiring them to the bridge and throwing on unknown ones. Adds the missing label, contact, and quick-reply APIs and tightens input validation to prevent accidental resets.

- **Bug Fixes**
  - Unknown `chatModify` variants now throw with the variant name.
  - Routed `disableLinkPreviews`, all label actions, quick replies, and `contact: null` → `removeContact`; fixed label arg order and pinned it to the bridge `.d.ts`.
  - Reject fields the wire can’t carry: `contact.lidJid`, `contact.pnJid`, `contact.username`, and `addLabel.predefinedId` (deletes unaffected).
  - Require both `name` and `color` on `addLabel` edits to avoid resetting missing fields.
  - Quick replies mint an id when `timestamp` is empty or missing.

- **New Features**
  - Added: `addLabel`, `addChatLabel`, `removeChatLabel`, `addMessageLabel`, `removeMessageLabel`, `addOrEditContact`, `removeContact`, `addOrEditQuickReply`, `removeQuickReply`, `star` (iterates per message).

<sup>Written for commit 11d150b4ca837831c8f7f9d13b9f678904458f72. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Added contact management, including adding, editing, and removing contacts.
  * Added chat and message labeling capabilities.
  * Added message starring and quick-reply management.
  * Added link-preview updates through chat modifications.

* **Bug Fixes**
  * Unsupported chat modifications now return clear errors instead of being ignored.
  * Improved validation and error handling for chat actions.

### DIFF
--- a/src/Socket/chat-actions.ts
+++ b/src/Socket/chat-actions.ts
@@ -96,7 +96,9 @@ export const makeChatActionMethods = (ctx: SocketContext) => {
 			} else if ('quickReply' in mod) {
 				// Upstream's `timestamp` is the app-state index key, the same slot
 				// the core calls `id`. Deleting is the same upsert with `deleted`.
-				const id = mod.quickReply.timestamp ?? String(Math.floor(Date.now() / 1000))
+				// `||`, not `??`: upstream treats an empty timestamp as absent and
+				// mints a key, and the core rejects an empty index outright.
+				const id = mod.quickReply.timestamp || String(Math.floor(Date.now() / 1000))
 				if (mod.quickReply.deleted) {
 					await client.deleteQuickReply(id)
 				} else {
--- a/src/Socket/chat-actions.ts
+++ b/src/Socket/chat-actions.ts
@@ -1,3 +1,6 @@
+import type { proto } from '../WAProto/runtime.ts'
+import type { QuickReplyAction } from '../Types/Bussines.ts'
+import type { LabelActionBody } from '../Types/Label.ts'
 import type { ChatModification, WAPatchName } from '../Types/index.ts'
 import { Boom } from '../Utils/boom.ts'
 import type { SocketContext } from './types.ts'
@@ -120,6 +123,51 @@ export const makeChatActionMethods = (ctx: SocketContext) => {
 		 */
 		resyncAppState: async (_collections?: readonly WAPatchName[], _isInitialSync?: boolean) => {
 			ctx.logger.info('resyncAppState: app state is synced automatically by the Rust bridge')
+		},
+
+		// Upstream models all of these as sugar over `chatModify`, and so do we:
+		// one place decides which bridge call a modification becomes, so the
+		// method and the modification can never disagree.
+
+		addOrEditContact: async (jid: string, contact: proto.SyncActionValue.IContactAction) => {
+			await methods.chatModify({ contact }, jid)
+		},
+
+		removeContact: async (jid: string) => {
+			await methods.chatModify({ contact: null }, jid)
+		},
+
+		addLabel: async (jid: string, labels: LabelActionBody) => {
+			await methods.chatModify({ addLabel: { ...labels } }, jid)
+		},
+
+		addChatLabel: async (jid: string, labelId: string) => {
+			await methods.chatModify({ addChatLabel: { labelId } }, jid)
+		},
+
+		removeChatLabel: async (jid: string, labelId: string) => {
+			await methods.chatModify({ removeChatLabel: { labelId } }, jid)
+		},
+
+		addMessageLabel: async (jid: string, messageId: string, labelId: string) => {
+			await methods.chatModify({ addMessageLabel: { labelId, messageId } }, jid)
+		},
+
+		removeMessageLabel: async (jid: string, messageId: string, labelId: string) => {
+			await methods.chatModify({ removeMessageLabel: { labelId, messageId } }, jid)
+		},
+
+		star: async (jid: string, messages: { id: string; fromMe?: boolean }[], star: boolean) => {
+			await methods.chatModify({ star: { messages, star } }, jid)
+		},
+
+		// No jid: a quick reply is account-wide, keyed only by its index.
+		addOrEditQuickReply: async (quickReply: QuickReplyAction) => {
+			await methods.chatModify({ quickReply }, '')
+		},
+
+		removeQuickReply: async (timestamp: string) => {
+			await methods.chatModify({ quickReply: { timestamp, deleted: true } }, '')
 		}
 	}
 	return methods
--- a/src/Socket/chat-actions.ts
+++ b/src/Socket/chat-actions.ts
@@ -5,6 +5,21 @@ import type { ChatModification, WAPatchName } from '../Types/index.ts'
 import { Boom } from '../Utils/boom.ts'
 import type { SocketContext } from './types.ts'
 
+/**
+ * Refuse a field the mutation carries but the wire call cannot. Dropping it
+ * would report success for a change that never happened, which is the failure
+ * this whole path exists to stop.
+ */
+const rejectUnsupportedFields = (where: string, value: object, fields: readonly string[]) => {
+	const present = fields.filter(field => (value as Record<string, unknown>)[field] != null)
+	if (present.length) {
+		throw new Boom(`${where}: ${present.join(', ')} cannot be set through this client`, {
+			statusCode: 400,
+			data: { fields: present }
+		})
+	}
+}
+
 export const makeChatActionMethods = (ctx: SocketContext) => {
 	const methods = {
 		pinChat: async (jid: string, pin: boolean) => {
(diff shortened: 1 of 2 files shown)


jlucaso1:
Valid, and fixed in `1711ea4`. `??` where upstream uses `||`.

The consequence is a bit sharper than "wrong key": the core rejects an empty app-state index outright (`quick reply id cannot be empty`), so `chatModify({ quickReply: { timestamp: '' } })` failed here against an input upstream accepts and mints a key for. Not silent, but still a divergence on a supported input.

Covered both the empty and the absent timestamp, and checked the test bites by reverting to `??`, which fails the empty case.

---

jlucaso1:
All three valid, all three fixed. Two of them were the same antipattern this PR is about, in miniature, which I had missed on my own input types.

**Dropped fields (`ea4c0eb`).** Confirmed against the core rather than the type alone: `create_label` builds `LabelEditAction { name, color, deleted }` and nothing else, and `saveContact` takes names plus the address-book flag while `IContactAction` also carries `lidJid`, `pnJid` and `username`. Both were accepted and discarded while the call resolved, so a caller setting a username got a successful promise and no username. They now reject naming the field. A label delete is unaffected, since that mutation carries no payload for `predefinedId` to be lost from.

**The harness only proved self-consistency (`5f72650`).** This was the best catch of the three and I had not seen it. Every label parameter is a string, so a swapped pair type checks; and since the fake accepts anything, a swap made in both the code and its test would agree with itself and stay green while mislabelling chats. The assertion was circular exactly where the risk is highest.

Rather than only documenting the limit, the suite now reads the parameter names out of the installed bridge declarations and asserts the order there, which is what `Bridge/__tests__/dts-drift.test.ts` already does for the event union. The two halves now mean something together: we pass the label first, and `label_id` is what the bridge declares first. I checked it bites by editing the bridge's own `.d.ts` to swap `addChatLabel` and rerunning, which fails.

`npm test` 703 passing, lint clean.

---

jlucaso1:
One valid and fixed, one that would have made us less compatible rather than more.

**The label edit was resetting fields (`11d150b`).** Correct: `LabelActionBody` makes `name` and `color` optional, the mutation replaces the whole label, so `{ id, name: 'VIP' }` reached the wire with colour zero and a colour-only edit cleared the name. Upstream can omit one because it builds the proto directly and an absent field stays absent; `createLabel` has no way to express that. Both are now required on a non-delete, rejecting rather than defaulting into a change nobody asked for. A delete is unaffected, since that mutation carries no payload.

**The quick reply case I am leaving as it is, and here is the evidence.** Upstream's own patch builder is:

```js
quickReplyAction: {
    count: 0,
    deleted: mod.quickReply.deleted || false,
    keywords: [],
    message: mod.quickReply.message || '',
    shortcut: mod.quickReply.shortcut || ''
}
```

`count` is hardcoded to `0` and `keywords` to `[]` on every edit, and `message`/`shortcut` fall back to the empty string exactly as they do here. So a partial quick reply edit is lossy in upstream too, and in one respect more so: it always clears keywords, while this passes through whatever the caller supplied.

The observation about the wire is right; the conclusion does not follow for a compatibility layer. Rejecting partial edits would diverge from upstream on an input upstream accepts, which is the failure mode this package exists to avoid. The one case where the wire genuinely bites, an empty `shortcut` or `message`, is caught by the core with a named error rather than being applied.

The difference from the label case is that upstream *can* send a partial label edit and cannot send a partial quick reply. So one is a divergence worth closing and the other is upstream's behaviour, matched.

`npm test` 705 passing, lint clean.

---

PR #35, opened 2026-08-08
fix(types): align the diverging public signatures with upstream

jlucaso1:
## Summary

The auditor reports 121 signature mismatches, which collapse to 58 distinct members once the socket's three aliases (`WASocket`, `default()`, `makeWASocket()`) are deduplicated. It compares signature strings, so that count mixes a parameter that was renamed, a shape that is merely different, and a shape that breaks the caller. Only the last is a bug.

All 58 are triaged below. **Two were real and are fixed**; both were argument-count divergences that stopped upstream code from compiling. One more is real and cannot be fixed in this layer, and is written up rather than papered over. The rest are noise or benign, and are documented so the next audit does not investigate them from scratch.

The criterion applied throughout, explicitly: *does a piece of code written against upstream's types compile and run against these?* If yes, benign. If no, real.

### Real, and fixed

| item | what broke | fix |
| --- | --- | --- |
| `downloadMediaMessage` | `ctx` was required where upstream has it optional, so upstream's three-argument call did not compile | `ctx` is optional, and without one the call uses the client the last `makeWASocket` registered, which is the fallback `downloadContentFromMessage` already offers for standalone calls. So the three-argument call downloads rather than merely explaining itself. It rejects only when neither a context nor a registration exists |
| `createParticipantNodes` | upstream's fourth argument `dsmMessage` was missing, so a four-argument call did not compile | accepted and refused, not accepted and dropped. Upstream uses it to encrypt a different plaintext for the caller's own other devices; the engine encrypts one payload for every recipient, so ignoring it would send those devices the wrong message |

### Real, and not fixable here

| item | why |
| --- | --- |
| the upload contract: `MediaGenerationOptions.upload`, `MessageContentGenerationOptions.upload`, `MessageGenerationOptions.upload`, `WASocket.waUploadToServer` | see below |

### Cascade

`generateWAMessage`, `generateWAMessageContent`, `prepareWAMessageMedia` and `sendMessage`'s `content` parameter all read as mismatches while their own top-level shapes match upstream exactly. Each is downstream of the upload contract or of a nested type, so there is nothing to fix on them directly. `handleIdentityChange` and `hasNonNullishProperty` are the same pattern, rooted in `IdentityChangeContext` and `AnyMessageContent`.

`IdentityChangeContext.debounceCache` and `.debounceCache.set` are a decision already recorded, not a new finding. The nominal cache type was removed deliberately, and this PR does not reopen it.

### Benign

| group | why upstream code still compiles and runs |
| --- | --- |
| `sendMessage`, `subscribeNewsletterUpdates` return types | this package returns `T` where upstream returns `T \| undefined` or `T \| null`. Stronger, and assignable to what a caller declared |
| every `newsletter*` return type | `NewsletterMetadata` where upstream has `unknown`. Also stronger, and an improvement that landed with the newsletter surface |
| `productCreate`, `productUpdate`, `productDelete`, `updateCoverPhoto` | return `never` because they refuse. `never` is assignable to whatever upstream declares |
| `generateThumbnail` | accepts `string \| Buffer` where upstream accepts `string`. Wider, so upstream calls are a subset |
| `toNumber` | accepts a structural `{ toNumber?, low?, high? }`, which upstream's `Long` satisfies |
| `fetchLatestBaileysVersion` | a discriminated union where upstream has one object with `isLatest: boolean`. Narrower, and assignable |
| `BufferJSON.replacer`, `.reviver` | `unknown` where upstream has `any`. `any` is assignable in both directions, and both still work as `JSON.stringify` and `JSON.parse` arguments |
| `addOrEditQuickReply` | `QuickReplyAction` where upstream has `any`. A well-formed quick reply still compiles; only passing junk stops |
| `getOrderDetails` | a third optional `sellerJid`. Upstream's two-argument call still compiles, and the argument is required at runtime for the reason recorded when it was added |
| `newsletterFetchMessages` | `since` and `after` optional where upstream has them required. More permissive |
| `resyncAppState` | optionality differs, deliberately. Belongs to the socket-internals PR and is not touched here |
| `WASocket.ws` and its emitter methods | the same type name on both sides, and the generic type parameter on `off`/`once`/`removeListener` and friends. Nominal identity, not a shape difference |
| `encodeBigEndian` | upstream's parameter names are minified to `e` and `t`. Renaming to match would make this code worse |
| `hkdf`, `md5` | re-exported from the bridge by both packages, so the names are the bridge's and not this layer's to change |
| `processHistoryMessage`, `downloadAndProcessHistorySyncNotification` | a named alias here against an inline object upstream, field for field the same shape. `pastParticipants` is required here and optional upstream, which constrains constructing the object but not reading it |

### Noise, left alone

`Curve.sign` (`buffer` vs `buf`), `Curve.verify` (`publicKey` vs `pubKey`), `relayMessage` (a named options object where upstream destructures in the signature), `downloadContentFromMessage` (`mediaContent` vs `__0`). TypeScript is structural: no caller names a positional parameter, so renaming these buys a smaller number in the report and nothing else. The two in `Curve` would also mean adopting names that are worse than the ones here.

One group of names was worth changing and is in its own commit, marked cosmetic so a reviewer can skip it: five refusal parameters carried underscore prefixes purely to satisfy the linter, and now carry upstream's names with the values discarded through `void`, which is the idiom the rest of this package already uses.

## Upload contract

The hypothesis was that one divergent upload contract caused a cascade of mismatches. Half right, and the half that is wrong changes what to do about it.

The exported `WAMediaUploadFunction` in this package is already character-for-character upstream's: `(encFilePath: string, opts: { fileEncSha256B64, mediaType, timeoutMs? })`. It is not the type of the `upload` field, and that is where the divergence lives. `MediaGenerationOptions.upload` here is `(media: Buffer, opts: { mediaType }) => Promise<UploadMediaResult>`.

The two are different in kind, not in spelling:

- upstream's uploader is a **transport**: encryption already happened, the ciphertext is on disk, hand back a URL
- this package's is **encryption plus transport**: here is plaintext, return the media key, both hashes and the length as well

So a consumer who passes their own upstream-shaped upload function gets a plaintext `Buffer` where they expect a file path, which is the silent-corruption case rather than a compile error. That is real, and it is the highest-impact item in this whole audit.

It is not fixable here. Accepting upstream's shape means encrypting the media in TypeScript, writing the ciphertext to a temp file and computing `fileEncSha256B64`, all beside the implementation the core already has. That is exactly the second implementation of protocol work the layer split exists to prevent, and it would be the wrong place for it.

**Gap, for the lower layers.** The core would need to expose encrypt-to-file, or encrypt-to-stream with the receipt, so this layer could hand a caller-supplied uploader an encrypted path and the hash upstream's contract promises. `agents_prompts/` does not exist in either the bridge or the core checkout, so this is recorded here rather than as a file: the capability wanted is "encrypt this media and give me back a path plus `fileEncSha256`, without uploading", after which the upstream `upload` contract becomes implementable in a follow-up.

Until then the divergence stays, and this PR does not disguise it.

## Breaking

Nothing that worked before stops working.

`downloadMediaMessage` accepts strictly more calls than it did: three arguments now compile and, after `makeWASocket` has initialized, download. `createParticipantNodes` also accepts strictly more: a fourth argument now compiles and rejects rather than failing to compile. Callers passing four arguments and expecting per-device plaintext substitution get an error, which is the point, since the previous alternative was for it not to compile at all.

One caveat is worth stating rather than burying: the client `downloadMediaMessage` falls back to is whichever socket was created last, so a host juggling several sockets should keep passing the context explicitly. That caveat is not new, it is the same one `downloadContentFromMessage` already carries, and it is documented where the fallback is.

The cosmetic commit renames parameters only. TypeScript does not bind positional arguments by name, so no call site moves.

## Left alone, and why

The benign and noise tables above are the full list, grouped. The one-line version: a shape that is stronger, wider or structurally compatible does not break a caller, and a parameter name is not part of the contract in a structural type system. Chasing either would grow the diff without changing what any consumer can do.

## Coverage

Audit against `baileys@7.0.0-rc13`:

- `main`: 97.71% (21843/22354), 121 function mismatches
- this branch: 97.73% (21846/22354), 118 function mismatches

Three entries moved, and that number is the honest measure of what fixing only the real items does. The mismatch counter does not go to zero and should not: most of what it counts is this package being deliberately stronger, wider or more explicit than upstream, and the upload contract will keep counting until the core can support it. A branch that zeroed this counter would have done so by giving up those choices.

## Validation

- `npm run format`, `npm run lint`, `tsc --noEmit`: clean
- `npm test`: 838 passing, 0 failing
- 5 new tests in `src/__tests__/signature-compatibility.test.ts`, covering `downloadMediaMessage` with an explicit context, with the registered client and no context, with a context present but empty, and with neither available, plus the four-argument `createParticipantNodes` call rejecting without sending anything
- Both fixes were checked by reverting the line they add and confirming the tests fail
- `test:e2e` not run, it needs a mock server not available here

### DIFF
--- a/src/Socket/business.ts
+++ b/src/Socket/business.ts
@@ -1,6 +1,7 @@
 import type { OrderResult, BusinessProfileUpdateInput } from '@oxidezap/whatsapp-rust-bridge'
-import type { CatalogPage, CollectionsPage, GetCatalogOptions } from '../Types/Product.ts'
+import type { CatalogPage, CollectionsPage, GetCatalogOptions, ProductCreate, ProductUpdate } from '../Types/Product.ts'
 import type { UpdateBussinesProfileProps } from '../Types/Bussines.ts'
+import type { WAMediaUpload } from '../Types/Message.ts'
 import { Boom } from '../Utils/boom.ts'
 import { jidNormalizedUser } from '../WABinary/jid-utils.ts'
 import type { SocketContext } from './types.ts'
@@ -11,7 +12,8 @@ import type { SocketContext } from './types.ts'
  * why instead of a bare TypeError, and rejecting because there is nothing to
  * call.
  */
-const noProductWriteRoute = (method: string): never => {
+const noProductWriteRoute = (method: string, ...ignored: unknown[]): never => {
+	void ignored
 	throw new Boom(
 		`${method} is not supported: the WhatsApp Web client has no product create, edit or delete operation, so there is nothing for this to call. Manage the catalog from the WhatsApp Business app.`,
 		{ statusCode: 501 }
@@ -112,7 +114,8 @@ export const makeBusinessMethods = (ctx: SocketContext) => ({
 	 * package's upload path cannot produce one: it requires a url and a
 	 * direct path, which that endpoint does not return.
 	 */
-	updateCoverPhoto: async (_photo: unknown): Promise<never> => {
+	updateCoverPhoto: async (photo: WAMediaUpload): Promise<never> => {
+		void photo
 		throw new Boom(
 			'updateCoverPhoto is not available yet: the cover photo upload returns an {fbid, meta_hmac, ts} receipt that this package cannot obtain. removeCoverPhoto works.',
 			{ statusCode: 501 }
@@ -123,9 +126,10 @@ export const makeBusinessMethods = (ctx: SocketContext) => ({
 		await (await ctx.getClient()).removeBusinessCoverPhoto(id)
 	},
 
-	productCreate: async (_create: unknown): Promise<never> => noProductWriteRoute('productCreate'),
+	productCreate: async (create: ProductCreate): Promise<never> => noProductWriteRoute('productCreate', create),
 
-	productUpdate: async (_productId: string, _update: unknown): Promise<never> => noProductWriteRoute('productUpdate'),
+	productUpdate: async (productId: string, update: ProductUpdate): Promise<never> =>
+		noProductWriteRoute('productUpdate', productId, update),
 
-	productDelete: async (_productIds: string[]): Promise<never> => noProductWriteRoute('productDelete')
+	productDelete: async (productIds: string[]): Promise<never> => noProductWriteRoute('productDelete', productIds)
 })
--- a/src/Socket/index.ts
+++ b/src/Socket/index.ts
@@ -857,11 +857,25 @@ const makeWASocket = (config: UserFacingSocketConfig) => {
 		sendRawMessage: async (data: Uint8Array | Buffer) => {
 			return (await ctx.getClient()).sendRawMessage(data instanceof Uint8Array ? data : new Uint8Array(data))
 		},
+		/**
+		 * `dsmMessage` is accepted so the signature matches upstream, and
+		 * refused rather than ignored. Upstream uses it to encrypt a different
+		 * plaintext for the caller's own other devices; the engine encrypts one
+		 * payload for every recipient, so honouring it is not possible here and
+		 * dropping it would send those devices the wrong message.
+		 */
 		createParticipantNodes: async (
 			jids: string[],
 			message: proto.IMessage,
-			extraAttrs?: BinaryNode['attrs']
+			extraAttrs?: BinaryNode['attrs'],
+			dsmMessage?: proto.IMessage
 		): Promise<{ nodes: BinaryNode[]; shouldIncludeDeviceIdentity: boolean }> => {
+			if (dsmMessage) {
+				throw new Boom(
+					'createParticipantNodes: dsmMessage is not supported, the engine encrypts one payload for every recipient and cannot substitute a different one for your own devices',
+					{ statusCode: 501 }
+				)
+			}
 			const bytes = encodeProto('Message', message as Record<string, unknown>)
 			return (await ctx.getClient()).createParticipantNodesBytes(jids, bytes, extraAttrs ?? {})
 		},
--- a/src/Socket/privacy.ts
+++ b/src/Socket/privacy.ts
@@ -63,7 +63,8 @@ export const makePrivacyMethods = (ctx: SocketContext) => {
 		 * Warned once rather than thrown: upstream callers await this inside a send
 		 * workflow, and rejecting would abort a workflow that was going to succeed.
 		 */
-		issuePrivacyTokens: async (jids: string[], _timestamp?: number): Promise<void> => {
+		issuePrivacyTokens: async (jids: string[], timestamp?: number): Promise<void> => {
+			void timestamp
 			if (!warnedAboutPrivacyTokens) {
 				warnedAboutPrivacyTokens = true
 				ctx.logger.warn(
--- a/src/Utils/messages.ts
+++ b/src/Utils/messages.ts
@@ -811,12 +811,19 @@ export const extractMessageContent = (content: WAMessageContent | undefined | nu
 	return content
 }
 
+/**
+ * Every field is optional, so upstream's `{ reuploadRequest, logger }` and this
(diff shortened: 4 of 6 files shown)


jlucaso1:
On "fall back to the registered client when context is omitted" in `src/Utils/messages.ts`: good catch, and it turns this into a real fix rather than a better-worded refusal. Done in `57a0ade`.

Making the parameter optional had only moved the failure from compile time to run time: the three-argument call compiled and then explained itself. `_registerActiveBridgeClient` already holds the client the last `makeWASocket` built, and `downloadContentFromMessage` uses exactly that fallback for standalone calls, so there was no reason for this one to insist on its own.

It now takes `ctx?.waClient ?? activeBridgeClient` and rejects only when neither exists, naming both ways to supply one. The multi-socket caveat carries over from the sibling function and is documented where the fallback is: the registration points at whichever client was created last, so a host juggling sockets should still pass the context.

Four tests: an explicit context is used, a three-argument call downloads through the registration, a context present but empty falls back rather than being dereferenced, and neither present rejects. Removing the fallback fails two of them.

The PR body's triage table is updated to match, since `downloadMediaMessage` is now a fix rather than a refusal.

---

jlucaso1:
On "make the fallback context type accept upstream callers" in `src/Utils/messages.ts`: valid on all three counts, fixed in `2aac251`. The third one is the sharpest, and I should have caught it.

Upstream's context is `{ reuploadRequest, logger }` with both required and no client at all, so an upstream caller passing it as written did not compile. My own error message told the caller to pass `{ waClient: sock.waClient }`, which did not compile either, because the other two fields were still required. And the tests used `as never`, which is exactly why neither showed up: a cast that silences the type is a test that stops testing the type.

Every field is optional now, which is what the runtime already assumed once the fallback existed. `reuploadRequest` and `logger` stay declared and are not read, and the type says so: the engine handles the re-upload request, the CDN failover and the logging, so a caller's versions would have nothing to do. Accepting and ignoring them is right here in a way it was not for `dsmMessage` on the other fix, because ignoring these changes no outcome, while ignoring that one would have sent the wrong message to the caller's own devices.

The casts are gone and a test now passes upstream's context shape verbatim. Restoring the required fields breaks the typecheck in three places in that file, which is the check that the casts were hiding.

---

jlucaso1:
On the `activeBridgeClient` fallback in `downloadMediaMessage`: the risk is real and stated correctly, and I am keeping the fallback. Worth laying out the tradeoff, because two reviewers have now argued opposite sides of it and the merge decision is the repo owner's, not mine.

The case against, which is yours: in a multi-account host the module-level pointer is whichever socket was created last, so a caller who omits `ctx` can have a download run against another session's client, with the wrong credentials and Signal pair. That fails cryptographically or reads the wrong account's media, and it does so silently.

The case for, which was the Codex finding earlier on this PR: without the fallback, upstream's documented three-argument call compiles and then always throws. That is not compatibility, it is a compile error moved to run time.

Three things tipped it toward keeping the fallback. The pattern is not new to the public surface: `downloadContentFromMessage` is exported and has had exactly this fallback, with the same caveat, since before this PR, so removing it here would leave the risk in place one function over while breaking the upstream call here. The failure is loud rather than quiet in practice, because the wrong client's Signal pair fails HMAC verification rather than returning someone else's plaintext. And single-account hosts, which are the common case, get an API that behaves as upstream's does.

What I have not done is pretend the risk away: the caveat is documented where the fallback is, in the same words as the sibling function, and the PR body calls it out under Breaking rather than burying it.

If you would rather this stayed confined to the internal helper, say so and I will require `ctx` here and revert the PR body's claim that upstream's three-argument call runs. The per-socket scoping you suggest is the better answer than either, but it is a change to how the registration works and belongs in its own PR rather than inside a signature-alignment one.

---

PR #37, opened 2026-08-08
feat(store): make the bridge keystore namespaces a documented contract

jlucaso1:
## Summary

A consumer writing a custom store gets two things it did not have:

- `isBridgeInternalKeyType(type)` and `BRIDGE_INTERNAL_KEY_TYPES`, exported from the package root, to tell bridge-internal state apart from Signal key material when enumerating, counting, exporting or migrating a store. Both derive from the routing catalog in `src/Compatibility/legacy-store/constants.ts`, not from a second list.
- A README section that says the key store receives bridge state, which prefixes it arrives under, that an empty key store is not a "no session" signal here, and what the first-run check actually is.

## The report

A production bot aliased to `@oxidezap/baileyrs@0.1.0` used its keys table to decide whether this was a first run:

```js
const stmtCountKeys = db.prepare('SELECT COUNT(*) AS n FROM wa_keys')
const isEmpty = () => !creds.registered && !creds.me?.id && stmtCountKeys.get().n === 0
```

After the migration the store reported a session that did not exist forever: the bot skipped its pairing-method menu, the lib emitted a QR nobody rendered, and the process sat in `connecting`. Dropping the `COUNT(*)` term fixed it.

Reproduced with an instrumented `SignalKeyStore`, fresh creds, a dead-end TCP listener so no pairing can happen:

```
creds.registered = false | creds.me = undefined

keystore set() calls before pairing: 8   (all to namespace bridge-native-device)
surviving rows: bridge-native-device / device

isEmpty() OLD => false   <- the defect
isEmpty() NEW => true
```

One detail worth recording: `bridge-native-device/account` is also written, as a null, so a store that treats a null write as a delete keeps one row while one that inserts nulls keeps two. Either way `COUNT(*)` is non-zero before pairing, which is all the broken check needed.

## Root cause

**Not pre-keys.** Pre-keys are uploaded at login in both upstream (`CB:success`) and here (`upload_pre_keys_at_login` in the core), so before pairing the keystore receives zero pre-key writes. That correction matters, because it changes what the fix is.

What is written is bridge-internal state. baileyrs uses the Baileys `SignalKeyStore` as the persistence channel for things that are not Signal keys, under the namespaces catalogued as `MirrorStore` (`bridge-native-*`, ten entries) and `RawStore` (`bridge-*`, seven). The first one to land is `bridge-native-device`: the core reads its device record during engine init, and `wrapLegacyStore`'s device route mirrors the projection back into the store on that read, while the socket is still `connecting`.

Upstream keeps the equivalent state in `creds`, written by `saveCreds` and kept out of `keys` entirely. That is the whole reason counting rows in `keys` works there and lies here.

Scope of who is affected: only the `wrapLegacyStore` path, i.e. the upstream `{ creds, keys }` shape (auto-wrapped, and `useLegacyMultiFileAuthState`). `useMultiFileAuthState` builds its `keys` as a projection over the engine's own store via `projectNativeStore`, which only ever speaks the Baileys namespaces, so no `bridge-` name surfaces there. The README says so, because it decides whether a reader needs the filter at all.

## Scope

Two fixes were possible:

- **A** separate the channels, so `keys` carries nothing but Signal key material and real parity is restored. Needs a migration for every session already paired against the current namespaces, including stores this project does not own.
- **B** treat the divergence as a public contract: documented, and filterable.

B is what this PR does. A stays the correct end state and is recorded as an open item in the local `COMPATIBILITY_TODO.md` (kept out of git via `.git/info/exclude`) with the deferral reason and the constraints a future migration has to respect. Nothing here renames, reshapes or reroutes an existing namespace, so no paired session is touched.

## The tests as documentation

`src/__tests__/bridge-keystore-namespaces.test.ts`:

- **The observed behavior, fixed.** A real socket, fresh creds, a local listener that accepts and drops every connection so pairing is impossible. It waits for the first keystore write, then asserts creds are still unpaired, every namespace written is classified internal, `bridge-native-device` is among them, and the reported `COUNT(*)` check is false while the creds check is true. Verified by mutation: making the adapter drop `bridge-`-prefixed buckets before `keys.set` makes this test time out on "the engine never wrote to the keystore" instead of passing.
- **Exhaustiveness over the catalog.** Iterates `MirrorStore` and `RawStore` and asserts each is internal, and separately asserts the exported list equals the catalog's `nativeType` set. Add `bridge-native-something` in six months and both keep holding; hardcode the list instead and the second one fails immediately (verified).
- **The other side.** Iterates `LegacyStore` plus `sender-key-memory` and asserts none is internal, so a classifier that answered "internal" to everything cannot pass.
- **The prefix trap.** `bridge`, `bridged-native-device`, `native-device`, `my-bridge-native-device`, `pre-key-bridge` are all not internal, while a catalog-absent `bridge-native-future-store` is, which pins the matching decision below.
- **Edges.** `''`, `undefined`, `null`, `0`, `42`, `{}`, `[]`, a symbol and a function all return false, none throws. Dropping the `typeof` guard fails this one (verified).

## Prefix matching

The classifier matches the reserved `bridge-` prefix first, then falls back to the derived exact set.

Prefix wins because the failure this PR exists to prevent is a *future* namespace surprising someone. A store filtering on an exact list has to be rebuilt and redeployed every time the catalog grows; the same store filtering on the prefix keeps working. The cost of a prefix match is misclassifying a legitimate Signal namespace that happens to start with `bridge-`, and there is none: no upstream `SignalDataTypeMap` key starts with it, so the prefix is declared reserved in the README.

The exact set is kept as a fallback rather than dropped, so a namespace that one day does not use the prefix is still classified, and it is exported in its own right because a SQL store wants it for a `NOT IN (...)` clause.

## Validation

```
npm run format
npm run lint          # tsc + oxlint, clean
npm test              # 846 tests, 217 suites, 0 failures
npm run compat:audit  # 97.92% (21888/22354), unchanged
```

The three new exports show up in the emitted declarations (`lib/Compatibility/legacy-store/namespaces.d.ts`, re-exported through `lib/Utils/wrap-legacy-store.d.ts`) and the auditor classifies all three as `extra`, not `missing` or `mismatch` -- upstream has no equivalent because it does not have the problem.

`test:e2e` not run: it needs a mock server that is not available here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GFh1ygyuQTkST9Lp5DmU76)_

---
## Summary by cubic
Formalize and document bridge keystore namespaces, and expose a classifier API so custom stores can ignore bridge-internal state. Prevents first-run checks from breaking when `bridge-native-device` is written before pairing.

- **New Features**
  - Exported `isBridgeInternalKeyType`, `BRIDGE_INTERNAL_KEY_TYPES`, and `BRIDGE_INTERNAL_KEY_PREFIX` from `@oxidezap/baileyrs` to classify and filter `bridge-*` namespaces. README adds “Bridge state in your key store”: explains that `bridge-*` keys are not Signal keys, can appear before pairing, the prefix is reserved, how to filter, and that backups/store-to-store moves must keep these rows. Also clarifies that `BRIDGE_INTERNAL_KEY_TYPES` lists only bridge-internal namespaces, and notes that aliased installs can import from `@whiskeysockets/baileys`.

- **Migration**
  - Do not use key counts to detect first run. Use `!creds.registered && !creds.me?.id`.
  - When producing an upstream-shaped view (enumerating/exporting to Baileys shape), filter with `isBridgeInternalKeyType(type)` or exclude `BRIDGE_INTERNAL_KEY_TYPES`. Do not drop `bridge-*` rows in backups or store-to-store moves.

<sup>Written for commit dfcbb7c42c1ee4dd1f0a9a7d8b1c88960f0e0fa8. Summary will update on new commits.</sup>


jlucaso1:
Two of the three Codex findings were real and are fixed in 048f0ed.

**NUL bytes (P2): correct, and worse than it looked.** The file had three literal NUL bytes at the `${type}\0${id}` delimiter, so git stored the blob as binary and `git show` reported nothing but "Binary files differ". Replaced with the `\0` escape, same runtime delimiter. The committed blob is text now, so the diff is reviewable.

**Backup and migration advice (P1): correct.** "Skip the bridge rows when exporting or migrating" was wrong advice for a full backup. The raw namespaces (`bridge-signed-prekey`, `bridge-sender-key-devices`, `bridge-base-key`, `bridge-sent-message`, `bridge-msg-secret`, `bridge-mutation-mac`, `bridge-meta`) have no `legacyType` in the routing catalog at all, and a session that took the `NATIVE_ONLY_PROJECTION` path is written to `bridge-native-session` with no legacy counterpart, so restoring only the non-bridge rows loses that state or rolls the session back. The advice is now scoped to upstream-shaped views and counts, with an explicit note that a backup or a baileyrs-to-baileyrs move carries every `bridge-` row across.

**Import specifier (P2): leaving as is.** Every example in this README imports from `@oxidezap/baileyrs`, and the drop-in section already states that an aliased install makes each `@whiskeysockets/baileys` import resolve here. Caveating this one snippet and no other would read as if the rule applied only to the classifier.

---

jlucaso1:
Both cubic findings taken, in dfcbb7c.

**`BRIDGE_INTERNAL_KEY_TYPES` label: right, the comment was wrong.** `wrapLegacyStore` writes the projected Baileys namespaces to the same store, so "everything the bridge can write" described the wrong set. Now says it is every bridge-internal namespace, and adds that the store also holds the namespaces the engine projects into it.

**Import specifier: reversing my earlier call.** I declined this when Codex raised it, on the grounds that every README example uses the canonical specifier and the alias mapping is stated once globally. Two reviewers landing on it changed my mind, and the argument that actually matters is one I missed: the Quick Start is written for new projects, but this section is written for people migrating a custom store off upstream, who are exactly the readers most likely to have installed under the alias and nothing else. One line above the import now points them at the specifier they have.

---

PR #38, opened 2026-08-09
fix(legacy-store): accept status/broadcast sender-key addresses

Salientekill:
## Problem

Sender keys are not exclusive to groups. Status updates and broadcast lists fan
out through the same sender-key mechanism, so `status@broadcast:<address>`
reaches the legacy-store translator alongside `<group>@g.us:<address>`.

`legacySenderKey` anchors the chat part on `@g.us:`:

```ts
const groupTerminator = `${SignalAddressSyntax.DOMAIN}${SignalDomain.GROUP}${SignalAddressSyntax.JID_DEVICE}` // "@g.us:"
const groupEnd = key.indexOf(groupTerminator)
if (groupEnd < 0) throw new TypeError(`invalid native sender-key address: ${key}`)
```

so every status address throws. Reproduction against `0.1.1`:

```js
import { legacyKey } from '@oxidezap/baileyrs/lib/Compatibility/legacy-store/routing.js'

legacyKey('sender_key', '120363000000000001@g.us:5511900000001@c.us.0') // ok
legacyKey('sender_key', 'status@broadcast:5511900000001:76@c.us.0')     // TypeError
```

Because the throw happens inside a storage operation, it aborts whatever
triggered it. In production this surfaced as `sendMessage` failing for an
**unrelated group** — the offending key has nothing to do with the chat being
written to:

```
Error [WhatsAppError]: storage operation failed: JS delete: invalid native
sender-key address: status@broadcast:5586XXXXXXXXX:76@c.us.0
  kind: 'storage',
  operation: 'JS delete: invalid native sender-key address: status@broadcast:5586XXXXXXXXX:76@c.us.0'
```

It reproduces on any bot that receives status updates while using an
upstream-style `auth: { creds, keys }` store (the `wrapLegacyStore` path).

## Fix

Parse the chat part by **shape** instead of by domain: the chat is everything
before the first `:` that follows the `@`, validated as a JID. Which chats fan
out through sender keys is the core's namespace, so this boundary validates the
form and stays out of that decision. Groups, `status@broadcast`, broadcast
lists and any chat kind the core adds later all translate without a new entry
here.

The same rule now applies in **both** directions. `nativeSenderKey` used to
accept any `<chat>::<user>::<device>` and hand back an address that
`legacySenderKey` would then reject:

```js
nativeKey('sender_key', 'nochat::5511900000001::0')      // ok -> nochat:5511900000001@c.us.0
legacyKey('sender_key', 'nochat:5511900000001@c.us.0')   // TypeError
```

That asymmetry was the evidence the original behaviour was a defect rather than
a policy, so closing it is part of the fix.

The shape check excludes whitespace and control characters, not just the
separators. No JID carries them, and one that slips through becomes a storage
key that is invisible in a log and unmatchable by a later lookup.

`constants.ts` is unchanged: `SignalDomain` is the vocabulary of Signal
*address* domains, and `broadcast` is a chat domain, so it does not belong
there.

## Tests

`src/Compatibility/__tests__/legacy-store-sender-key-routing.test.ts` covers
group, status and broadcast addresses, an unknown chat domain, chats carrying
whitespace or a control character, the round-trip in both directions, and keeps
the rejection for malformed addresses.

Batch behaviour is deliberately unchanged: a genuinely malformed key still
fails the whole `deleteMany`, so a bad key surfaces instead of being skipped in
silence.

- `node --test` — 909 passing
- `tsc --noEmit` — clean
- `oxfmt --check src` / `oxlint` — clean

Each fix commit is paired with the test that pins it, and reverting the fix
alone makes that test fail — so the tests are tied to the behaviour rather than
to the implementation.

---
## Summary by cubic
Accept sender-key addresses from status, broadcast lists, and any chat JID in the legacy-store translator to stop storage errors during status reception. This prevents unrelated `sendMessage` calls from failing.

- **Bug Fixes**
  - Parse chat by JID shape (not domain) so groups, `status@broadcast`, broadcast lists, and future chat kinds translate.
  - Make `legacyKey` and `nativeKey` symmetric; enforce a valid chat JID (reject whitespace/control chars); reject malformed addresses in both directions; add tests for unknown domains and round-trips.
  - Fix batch deletes: mixed group/status `deleteMany` succeeds; malformed keys still fail the batch.

<sup>Written for commit c9b9587d5de6fab31c73b0cc00923452f1bf58ed. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sender-key routing for status updates and broadcast lists.
  * Added validation for supported chat address formats.
  * Ensured sender-key conversions reject malformed addresses with clear errors.
  * Improved batch deletion so valid keys are removed safely while malformed entries are rejected without partial deletion.

* **Tests**
  * Added coverage for group, status, broadcast, hosted, unknown, and malformed chat addresses.

### DIFF
--- a/src/Compatibility/__tests__/legacy-store-sender-key-routing.test.ts
+++ b/src/Compatibility/__tests__/legacy-store-sender-key-routing.test.ts
@@ -1,12 +1,9 @@
 /**
- * Sender-key addresses are not exclusive to groups.
- *
- * Status updates and broadcast lists fan out through the same sender-key
- * mechanism as groups, so `status@broadcast:<address>` reaches the legacy-store
- * translator just like `<group>@g.us:<address>` does. Anchoring the chat part on
- * `@g.us:` made every status address throw `invalid native sender-key address`,
- * and since the throw surfaces inside a storage operation it aborts whatever
- * triggered it — a `sendMessage` to an unrelated group, for example.
+ * A sender key is `<chatJid>:<signalAddress>`, and the chat is not always a
+ * group: status updates and broadcast lists fan out through the very same
+ * mechanism, so `status@broadcast:<address>` reaches this translator too.
+ * The throw lands inside a storage operation, so it aborts whatever triggered
+ * it — a welcome message to an unrelated group, for example.
  */
 
 import { describe, it } from 'node:test'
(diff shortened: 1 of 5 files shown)


jlucaso1:
@codex review

PR #43, opened 2026-08-10
test(fuzz): differential fuzzing against upstream Baileys

jlucaso1:
Adds a differential and property-based fuzz suite that generates its own inputs and compares baileyrs against upstream Baileys directly.

## Why

The repository already compares itself to Baileys three ways — the declaration audit, the wire-fidelity audit, and ~50 behavioural compatibility suites — and all three run on inputs somebody wrote down. They find what we imagined. This adds the layer that generates them, using the assets already in the repo as grammars: `PROTO_MESSAGE_SCHEMAS` (498 message types), the `Bridge/schema.ts` event table (58 variants), and a hand-written JID grammar.

The runs recorded **23 differences: 20 open, 3 deliberate**, each in `src/__fuzz__/harness/divergence.ts` with a reason and a review date, most with a minimised reproducer in `src/__fuzz__/corpus/`.

**Several of the 20 are rediscoveries, not discoveries.** `KNOWN_WIRE_GAPS` and `KNOWN_UNSUPPORTED_CODECS` in `scripts/compatibility/proto-runtime-audit.ts` already track the proto schema gaps below, and `scripts/compatibility/__tests__/wire-fidelity.test.ts` already pins the field-number one by name. They are split out under their own heading rather than mixed into the new findings — a registry that presents known gaps as news is a registry nobody should trust about the entries that *are* new.

## New findings

Ordered by how much they matter. None are fixed here — each is a behaviour change to a published API and deserves its own commit and its own review.

**The bridge decoder throws on any 64-bit field outside ±(2^53−1).** Not precision loss: `decodeProto` raises `Value is larger than Number.MAX_SAFE_INTEGER` and the whole message fails to decode, where protobufjs returns a Long. The boundary is exact — `9007199254740991` decodes, `9007199254740992` throws. A legitimate server payload with a large `fileLength` or a microsecond timestamp becomes an error instead of a message. It also means baileyrs cannot always read back bytes it just handed the bridge.

**The bridge event adapter table is a plain object indexed by an untrusted string.** An event type of `constructor` or `toString` resolves through `Object.prototype`, the inherited function is called, and its return value is handed on as a canonical event; `__proto__` throws `adapter is not a function`. The type comes from the server by way of the runtime. A `Map`, `Object.create(null)`, or an `Object.hasOwn` guard fixes it.

[...]

PR #47, opened 2026-08-10
fix: three parity defects the differential fuzzers found

jlucaso1:
Three defects the fuzzers from #43 left open in the registry. Two in the bridge's event dispatch, one in the send path.

## 1. A missing data slot threw instead of dropping the event

30 of the 58 declared event types threw a `TypeError` when an event arrived with no data slot — `data.code` on `undefined`. `adapt.ts` documents the opposite contract ("Result is `null` on unrecoverable shape mismatch"), and the throw does not stay local: it propagates into the socket's event dispatch and takes down the whole event loop rather than the one event.

Every adapter in the table now reads its slot optionally, so the dispatch passes it through exactly as it arrived. Across all 58 types, with both `undefined` and `null`:

```
0 throw · 26 null · 21 noop · 11 emit
```

Six of the eleven never read a slot at all; the other five return a fallback their own author already wrote (`stream_error` its `?? 'unknown'`, `pair_error` its `'Unknown pairing error'`, `offline_sync_completed` its `count: 0`).

> **An approach that was tried and reverted.** The first version substituted `{}` for a missing slot at the dispatch instead. That removes the throws, but `{}` is a perfectly good object — so it walks straight past the `isObject(data)` guard several adapters use to decide a payload is unrecoverable. `history_sync` then built an empty final batch and the socket emitted a spurious `messaging-history.set` where it had previously dropped the event. Caught in review of this PR; the dispatch must not pre-empt that decision, so each adapter owns it. Pinned.

## 2. The event type resolved through the prototype chain

The adapter table is a plain object literal indexed by a string the runtime takes from the server, so an untrusted value was indexing a prototype-bearing lookup. `constructor` and `toString` resolved to inherited functions, were **called**, and had their return values handed on as canonical events — an object and a string respectively. `__proto__` and `valueOf` reached the other failure mode and threw.

The lookup now requires an own property.

```
                          before                  after
data: undefined     30 types throw          0 types throw
data: null          30 types throw          0 types throw
type "constructor"  leaked an object        null (dropped)
type "toString"     leaked a string         null (dropped)
type "__proto__"    threw                   null (dropped)
type "valueOf"      threw                   null (dropped)
```

Events that were already fine are untouched: `connected` still adapts, `qr` with a code still adapts, `qr` without one still drops.

## 3. Forwarding rewrote the caller's own message

`generateForwardMessageContent` shallow-cloned the outer message map and then assigned `contextInfo` onto the *nested* object, which that clone still shares with the caller. And it replaced rather than merged, so forwarding a quoted message stripped `stanzaId` and `participant` from the caller's own object. Upstream leaves the argument untouched.

Only the one object being written is rebuilt, rather than buying upstream's `proto.Message.decode(proto.Message.encode(content))` round trip on a send path. Allocation count is unchanged on the plain-text path — the `extendedTextMessage` built there is already ours, so the context goes straight into the literal — and one shallow spread higher elsewhere, which is the irreducible cost of not writing into someone else's object.

Benchmarked with each input consumed exactly once, because the old implementation mutates its argument and a reused pool would have it take a different path from the second pass onward (the forwarding score accumulates):

| shape | before | after |
|---|---|---|
| conversation (plain text) | 540–556 ns/op | 530–538 ns/op |
| extendedTextMessage | 221–253 ns/op | 155–188 ns/op |
| imageMessage (8 fields) | 228–302 ns/op | 198–201 ns/op |

**Two of the three registry entries on this helper stay open, deliberately.** `copy-shape` and `int64-truncation` come from upstream's round trip converting through `Long` and reshaping the key set — not from the aliasing — so matching them means buying exactly the round trip this avoids. The fuzzer settles which is which rather than a reading of the code: after the fix those two still excuse findings and the mutation entry excuses nothing.

## Registry

Three entries deleted, along with the helpers only they used: `bridge-adapter-throws-on-missing-data`, `bridge-adapter-prototype-chain-lookup`, `forward-message-content-mutates-input`. An allowlist that outlives its divergence is how the same bug comes back unnoticed. Open findings go from 25 to 22.

## Verification

New pins in `src/Bridge/__tests__/adapt.test.ts` and `src/__tests__/forward-message-content.test.ts`, each of which fails when its fix is removed:

```
prototype guard removed        -> # fail 2
data normalisation removed     -> # fail 2
forwarding aliasing restored   -> # fail 3
```

`tsc --noEmit`, `oxlint`, `oxfmt --check` clean. `npm test` 1181 pass / 0 fail. `npm run fuzz` 185 pass / 0 fail. Deep `pure-differential` at 90 pass / 2 fail, matching the pre-change baseline exactly; bridge fuzzer clean on seeds `fixed d1 d5 d9 y1`.

## Not in this change

`proto-decode-above-max-safe-integer` is the most severe finding in the registry — any 64-bit field outside ±(2^53−1) fails the *whole message* decode — but the throw comes from `@oxidezap/whatsapp-rust-bridge`, so it needs a change there rather than here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01RvtvVVWQs8AeBqSzS1JpCg

### DIFF
--- a/src/Bridge/__tests__/adapt.test.ts
+++ b/src/Bridge/__tests__/adapt.test.ts
@@ -811,6 +811,20 @@ describe('adaptBridgeEvent — anti-corruption layer', () => {
 			}
 		})
 
+		/**
+		 * The dispatch must not answer this question on an adapter's behalf.
+		 *
+		 * Substituting `{}` for a missing slot was tried and reverted: `{}` is a
+		 * perfectly good object, so it walks past the `isObject(data)` guard several
+		 * adapters use to decide a payload is unrecoverable. `history_sync` then
+		 * built an empty final batch and the socket emitted a spurious
+		 * `messaging-history.set` where it had previously dropped the event.
+		 */
+		it("leaves an adapter's own object guard the last word", () => {
+			expect(adaptBridgeEvent({ type: 'history_sync' } as never)).toBe(null)
+			expect(adaptBridgeEvent({ type: 'history_sync', data: null } as never)).toBe(null)
+		})
+
 		it('drops the events that need their data, and keeps the ones that do not', () => {
 			// `qr` reads a code it has not been given, so there is nothing to emit.
 			expect(adaptBridgeEvent({ type: 'qr' } as never)).toBe(null)
--- a/src/Bridge/__tests__/adapt.test.ts
+++ b/src/Bridge/__tests__/adapt.test.ts
@@ -1,6 +1,6 @@
 import { describe, it } from 'node:test'
 import type { MessageWireInfo, WhatsAppEvent } from '@oxidezap/whatsapp-rust-bridge'
-import { adaptBridgeEvent, adaptBridgeMessageWire } from '../adapt.ts'
+import { adaptBridgeEvent, adaptBridgeMessageWire, KNOWN_BRIDGE_EVENT_TYPES } from '../adapt.ts'
 import { expect } from '../../__tests__/expect.ts'
 import {
 	groupAnnouncementWireFixture,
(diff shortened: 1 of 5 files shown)


jlucaso1:
Two commits pushed. The review finding was a real regression in my previous commit, and the PR now also carries a second fix.

## The `history_sync` finding — correct, and a regression I introduced (`8246bd1`)

Substituting `{}` for a missing data slot removed the throws, but `{}` is a perfectly good object, so it walked straight past the `isObject(data)` guard several adapters use to decide a payload is unrecoverable. `history_sync` built an empty final batch and the socket emitted a spurious `messaging-history.set` where it had previously dropped the event. My own probe had shown it and I read past it:

```
history_sync -> {"type":"historySync","chats":[],"contacts":[],"messages":[], …}
```

The dispatch no longer pre-empts that decision. The slot is passed through as it arrived and every adapter in the table now reads it optionally, which keeps the "never throw" property while leaving each adapter the last word on its own payload. Across all 58 declared types, with both `undefined` and `null`:

```
0 throw · 26 null · 21 noop · 11 emit
emit: connected, disconnected, stream_replaced, client_outdated,
      temporary_ban, qr_scanned_without_multidevice, logged_out,
      pair_error, connect_failure, stream_error, offline_sync_completed
```

Six of those eleven never read a slot at all; the other five return a fallback their own author wrote. `history_sync` is back to `null` and pinned.

## Second fix: forwarding no longer rewrites the caller's message (`340aeed`)

`generateForwardMessageContent` shallow-cloned the outer map and then assigned `contextInfo` onto the *nested* object, which the clone still shares with the caller — and it replaced rather than merged, so forwarding a quoted message stripped `stanzaId` and `participant` from the caller's own object.

Only the one object being written is rebuilt, rather than buying upstream's `proto.Message.decode(proto.Message.encode(content))` round trip on a send path. Allocation count is unchanged on the plain-text path and one shallow spread higher elsewhere — the irreducible cost of not writing into someone else's object.

Benchmarked with each input consumed once, because the old implementation mutates its argument and a reused pool would have it take a different path from the second pass on (the forwarding score accumulates). My first benchmark did reuse a pool and was therefore meaningless:

| shape | before | after |
|---|---|---|
| conversation | 540–556 ns/op | 530–538 ns/op |
| extendedTextMessage | 221–253 ns/op | 155–188 ns/op |
| imageMessage (8 fields) | 228–302 ns/op | 198–201 ns/op |

**Two of the three registry entries on this helper stay open, deliberately.** `copy-shape` and `int64-truncation` come from upstream's round trip converting through `Long` and reshaping the key set — not from the aliasing — so matching them means buying that round trip. The fuzzer settles it rather than my reading: after the fix those two still excuse findings and the mutation entry excuses nothing, so only that one is deleted. I had earlier guessed this change would close all three; it closes one.

An earlier version of the fix spread the nested slot unconditionally and invented a property named `"undefined"` on a message with no content key, where upstream raises a `TypeError`. The differential caught it (`89 pass / 3 fail` against a `90 / 2` baseline); that path now keeps the original assignment.

## Verification

`tsc --noEmit`, `oxlint`, `oxfmt --check` clean. `npm test` 1181 pass / 0 fail. `npm run fuzz` 185 pass / 0 fail. Deep `pure-differential` back to `90 pass / 2 fail`, matching the pre-change baseline exactly. New pins fail when their fix is removed: 3 failures for the forwarding aliasing, 2 for the bridge guards.

---

PR #59, opened 2026-08-12
fix(bridge): read the payload shapes the bridge actually sends

jlucaso1:
## Summary

Five fields reached consumers as `undefined` while the server had supplied them. Each one is read with a strict number guard in the adapters, and none of them crosses the bridge as a number.

| field | crosses as | why |
|---|---|---|
| `presence.last_seen` | RFC 3339 string | `Option<DateTime<Utc>>` with no chrono `ts_*` module |
| `server_ack.timestamp` | RFC 3339 string | same |
| `pin_update.timestamp` | RFC 3339 string | same |
| `mute_update.timestamp` | RFC 3339 string | same |
| `mute_update` → `muteEndTimestamp` | protobufjs `Long` | a protobuf `int64` does not fit a JS number |
| `temporary_ban.expire` | `{ secs, nanos }` | serde's shape for `std::time::Duration` |

`dirty_state.timestamp` is a `u64` in the core, crosses as a number, and is left alone — with a test pinning that it stays that way.

## How each one got here

The four `DateTime` ones have been wrong for as long as those adapters have existed. `toUnixSeconds`, which already accepts both spellings, is used elsewhere in the same file; these four reached for `asNumber` instead.

`muteEndTimestamp` is a **regression from `whatsapp-rust-bridge` 0.10.0**. [whatsapp-rust-bridge#54](https://github.com/oxidezap/whatsapp-rust-bridge/pull/54) moved an app-state mutation onto the proto serializer so its keys match the declaration they were always published under — the right fix — and a side effect is that a 64-bit field stops being a plain number and becomes `{ low, high, unsigned }`. Measured against the adapter as it stood:

```
before 0.10.0 (serde, snake_case + number)  -> muteEndTimestamp = 1699999999
after  0.10.0 (proto, camelCase + Long)     -> muteEndTimestamp = undefined
```

The camelCase half of that change costs nothing here: these adapters already read both spellings.

`temporary_ban.expire` is a mismatch on the bridge side that this absorbs rather than waits on. The generated declaration maps `Duration` to `number`, while plain serde writes `{ secs, nanos }`, so a consumer reading the declared type gets an object where it expected a count. Worth raising there; not worth leaving broken here in the meantime.

## Design

Both forms are accepted for every field — today's and the one before it. These payloads have already changed shape once, and absorbing that is what this layer is for; pinning only the current form is how the same bug comes back on the next release. The new helpers live beside the existing guards in `Bridge/primitives.ts`:

- `asUnixSeconds` — like `toUnixSeconds`, but absence stays absent instead of becoming the epoch. These are optional fields: missing means "the server did not say", which is a different claim from 1970.
- `asInt64` — a `Long`, via its own `toNumber` when it has one, otherwise recombined from `low`/`high`. Only when `high` is `0` or `-1`, the range a JS number represents exactly; past that the value is dropped rather than silently rounded, because a wrong instant is worse than a missing one.
- `asDurationSeconds` — `{ secs, nanos }` or a plain count, to whole seconds. Every caller is a ban or a backoff measured in seconds.

## Validation

`npm test` — 1238 passing, no existing test needed adapting. `tsc`, `oxlint`, `oxfmt` clean.

New: `src/Bridge/__tests__/payload-shapes.test.ts`, 14 tests. Each field is pinned in both forms, plus absence stays absent, an unparseable timestamp is dropped rather than guessed, a `Long` past 2^53 is dropped rather than rounded, and `dirty_state` keeps crossing as a number.

Each helper was verified to be load-bearing by reverting it to `asNumber` and re-running:

| reverted | tests failing |
|---|---|
| `asUnixSeconds` | 4 |
| `asInt64` | 2 |
| `asDurationSeconds` | 1 |

---
## Summary by cubic
Accepts bridge payloads as sent and reports temporary bans as deadlines. Old behavior: strict number guards dropped RFC3339 timestamps, `Long` int64s, and `{ secs, nanos }` durations, and passed `temporary_ban.expire` through as a relative duration; new behavior: accepts those shapes (and prior numeric forms), converts bans to a unix-seconds deadline, and rejects invalid RFC3339 (including impossible calendar days and 24:00) and unsafe int64s, as well as deadlines a `Date` cannot represent.

- Uses `asUnixSeconds` for `presence.last_seen`, `server_ack.timestamp`, `pin_update.timestamp`, and `mute_update.timestamp`; uses `asInt64` for `muteEndTimestamp`; converts `temporary_ban.expire` via `absoluteFromDuration(asDurationSeconds(...))`; keeps `dirty_state.timestamp` as a number.
- Adds `asUnixSeconds` (RFC3339-only, refuses impossible days and 24:00), `asInt64` (recombines signed halves, requires exact safe integers, rejects rounded `toNumber`), `asDurationSeconds` (accepts `{ secs, nanos }` or number), and `absoluteFromDuration` (guards `Date` range).
- Compatibility: accepts current and legacy shapes; optional timestamps remain undefined.
- Migration: if you treated `temporary_ban.expire` as a relative duration, update callers to read a unix-seconds deadline.
- Tests: adds `Bridge/__tests__/payload-shapes.test.ts`; updates `temporary_ban` checks in `src/__tests__/regressions.test.ts` to assert a deadline window.

<sup>Written for commit f03e2babe364a29e9a63a95b3e140f9f07b47686. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of timestamps, durations, and 64-bit numeric values across bridge data.
  * Preserved missing or invalid values instead of incorrectly converting them to zero or epoch time.
  * Added support for multiple timestamp and field formats, including structured durations and legacy numeric representations.
  * Prevented unsafe large numeric values from being misinterpreted.

### DIFF
--- a/src/Bridge/__tests__/payload-shapes.test.ts
+++ b/src/Bridge/__tests__/payload-shapes.test.ts
@@ -99,6 +99,22 @@ describe('bridge payload shapes', () => {
 			}
 		})
 
+		it('refuses the 24:00 rollover, which RFC 3339 does not allow', () => {
+			// `Date.parse('2023-11-14T24:00:00Z')` is midnight on the 15th, so this
+			// is the other component that moves the instant to another day.
+			const rollover = adapt({
+				type: 'presence',
+				data: { from: JID, unavailable: false, last_seen: '2023-11-14T24:00:00Z' }
+			})
+			expect((rollover as { lastSeen?: number }).lastSeen).toBeUndefined()
+			// The last instant the hour does allow still parses.
+			const last = adapt({
+				type: 'presence',
+				data: { from: JID, unavailable: false, last_seen: '2023-11-14T23:59:59Z' }
+			})
+			expect(typeof (last as { lastSeen?: number }).lastSeen).toBe('number')
+		})
+
 		it('keeps the leap days that do exist, centuries included', () => {
 			// 2000 is a leap year and 1900 is not, which is the rule a plain
 			// `year % 4` check gets wrong.
@@ -228,6 +244,21 @@ describe('bridge payload shapes', () => {
 			const check = expectedDeadline()
 			const plain = adapt({ type: 'temporary_ban', data: { code: 101, expire: HOUR } })
 			check((plain as { expire?: number }).expire)
+		})
+
+		it('refuses a deadline a Date cannot hold', () => {
+			// The socket formats this with `new Date(expire * 1000).toISOString()`,
+			// which throws past ±8.64e15 ms — and that throw would land inside the
+			// event dispatch. No deadline beats one that takes the dispatch down.
+			const absurd = adapt({
+				type: 'temporary_ban',
+				data: { code: 101, expire: { secs: 8_640_000_000_000, nanos: 0 } }
+			})
+			expect((absurd as { expire?: number }).expire).toBeUndefined()
+			// And the socket's own formatting stays safe on what does get through.
+			const ban = adapt({ type: 'temporary_ban', data: { code: 101, expire: HOUR } })
+			const expire = (ban as { expire?: number }).expire!
+			expect(() => new Date(expire * 1000).toISOString()).not.toThrow()
 		})
 
 		it('an absent expire stays absent rather than becoming right now', () => {
--- a/src/Bridge/__tests__/payload-shapes.test.ts
+++ b/src/Bridge/__tests__/payload-shapes.test.ts
@@ -58,11 +58,15 @@ describe('bridge payload shapes', () => {
 		it('pin_update: timestamp', () => {
 			const iso = adapt({ type: 'pin_update', data: { jid: JID, timestamp: ISO, action: { pinned: true } } })
 			expect(iso).toMatchObject({ type: 'pinUpdate', timestamp: SECONDS, pinned: true })
+			const seconds = adapt({ type: 'pin_update', data: { jid: JID, timestamp: SECONDS, action: { pinned: true } } })
+			expect(seconds).toMatchObject({ type: 'pinUpdate', timestamp: SECONDS, pinned: true })
 		})
 
 		it('mute_update: timestamp', () => {
 			const iso = adapt({ type: 'mute_update', data: { jid: JID, timestamp: ISO, action: { muted: true } } })
 			expect(iso).toMatchObject({ type: 'muteUpdate', timestamp: SECONDS, muted: true })
+			const seconds = adapt({ type: 'mute_update', data: { jid: JID, timestamp: SECONDS, action: { muted: true } } })
+			expect(seconds).toMatchObject({ type: 'muteUpdate', timestamp: SECONDS, muted: true })
 		})
 
 		it('an absent timestamp stays absent rather than becoming the epoch', () => {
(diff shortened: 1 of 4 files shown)


jlucaso1:
All four addressed.

**The ban duration (P1) was the real one.** The wire states how long a ban lasts, the canonical event promises the instant it lifts, and `Socket/events.ts` formats it with `new Date(expire * 1000)` while `Example/example.ts` subtracts `Date.now()` from it. Passing the duration through would have put every ban in 1970 and let a bot retry immediately. Converted in the adapter, once. Two existing `temporary_ban` tests changed with it — that is the contract change, not an adaptation of convenience, and both now assert the deadline lands inside the window the call spans.

**`asInt64` was wrong on the sign, not just on the range.** Reading `low`/`high` separately gets it right by accident: `{ low: 0, high: -1 }` is -2^32 and it reported 0. It recombines the pair now. And `Long.toNumber` rounds rather than refusing past 2^53, so that path is checked with `Number.isSafeInteger` too, as is a plain number.

**RFC 3339 is matched before parsing.** `Date.parse('0')` is the year 2000, so a numeric-string sentinel became a believable instant. Applied to `toUnixSeconds` as well, not only the new helper — same hazard, same file.

**Numeric timestamp cases added** for `pin_update` and `mute_update`.

On cost, since this touches parsing: the string branch is the only one that gained work, and it carries app-state and presence traffic. Anything that crosses as a number — every message and receipt — returns before the regex. Measured at 1.28 ns/op for the number path against 85 ns/op for the string one, where the `Date.parse` that was already there dominates and the regex adds about 8 ns. The pattern is a module constant, so it is compiled once.

1248 passing.

jlucaso1:
Confirmed and fixed. `Date.parse` refuses month 13, hour 25 and second 61, but rolls a day past the end of its month forward — `2023-02-30T12:00:00Z` is March 2, `2023-11-31` is December 1 — so that was the one component still able to move an instant by days and look believable. The day is now checked against its month, with Gregorian leap years: 2000 has a February 29, 1900 does not.

Measured, since this sits in the parse path. Reading the components with capture groups doubled the string branch, 180 ns/op against 85 before, so the pattern stays a `test` and year, month and day come from fixed offsets instead — the format guarantees them. That lands at 97 ns/op, meaning the whole calendar check costs about 12 ns, and a day that cannot exist is refused at 16 ns/op without reaching `Date.parse`. A number still returns before any of it, at 1.30 ns/op.

1250 passing.

jlucaso1:
Both confirmed and fixed.

`2023-11-14T24:00:00Z` parses as midnight on the 15th — same class as the day check, an instant moved to another date. RFC 3339 stops the hour at 23, so that is checked now.

The oversized deadline is the worse of the two: `Socket/events.ts` formats a ban with `new Date(evt.expire * 1000).toISOString()`, which throws a RangeError past ±8.64e15 ms, and that throw would land inside the event dispatch rather than in the consumer. A duration that cannot produce a representable instant now reports nothing — no deadline beats one that takes the dispatch down. There is a test asserting the socket's own formatting stays safe on what does get through.

Cost is flat: one more fixed-offset read, 87 ns/op on the string path against 97 before, an hour past 23 refused at 16 ns/op without reaching `Date.parse`, and a number still returning at 1.28 ns/op.

1252 passing.

PR #61, opened 2026-08-12
fix(send): keep a caller's <biz> working, on bridge 0.11.0

jlucaso1:
## Summary

Bridge `0.10.0` → `0.11.0`, which carries [whatsapp-rust#1296](https://github.com/oxidezap/whatsapp-rust/pull/1296), plus the compatibility work that bump requires here.

An interactive message with a native-flow button gets a `<biz>` derived from its content. A caller that supplies its own put two on one `<message>`, and the client silently declines to render that — the server acks, delivered and read receipts come back, and the button appears nowhere. The engine now refuses the caller's node instead of stacking both, which is right: the two can disagree on `native_flow_name` and nothing at that layer knows which the server honours.

That leaves this library with a problem to absorb. Upstream Baileys derives nothing, so **attaching that node is what every payment recipe in that ecosystem tells a caller to do** — and that code is exactly what runs here. Left alone, code that works upstream would start throwing.

Found from a live bot whose PIX button stopped rendering. Measured both ways on the same account and group:

```xml
<!-- what the bot sent: nothing renders -->
<biz actual_actors="2" host_storage="2" native_flow_name="payment_info" privacy_mode_ts="…"/>
<biz native_flow_name="payment_info"/>

<!-- derived node alone: renders -->
<biz actual_actors="2" host_storage="2" native_flow_name="payment_info" privacy_mode_ts="…"/>
```

## Design

`relayMessage` retries once, without the tag the engine named, when the engine refuses for that reason.

**Retried rather than filtered up front.** Only the engine knows which nodes it derives for a given message. Replicating that rule here — "does this carry an interactive payload with buttons or a renderable envelope" — would duplicate core logic and go out of sync with the first button WhatsApp adds. The refusal happens while the engine validates, before it encrypts or sends anything, so the retry cannot deliver the message twice.

**Kept narrow.** The tag is read from the rejection's own sentence (`extra stanza child <biz> conflicts with the one this send derives`), so an unrelated `invalid-argument` — a malformed JID, an out-of-range enum — still reaches the caller untouched. A blanket retry on `invalid-argument` is how one send becomes two messages, and that case has its own test.

**The escape hatch stays open.** A caller node the engine did not derive is never dropped, because on a message that derives nothing the caller's node is the only one there is. The engine preserves that too, and this matches it rather than second-guessing it.

**Cost:** nothing is scanned on a send that succeeds. The work exists only after a refusal that would otherwise have been an error.

## Alternatives rejected

- **Filter `<biz>` unconditionally at the boundary** — breaks the escape hatch above.
- **Replicate the inference rule** — duplicates core logic, drifts on the next button name.
- **Ask the engine whether it will derive** — needs new core API for something the retry already answers.

## Validation

`npm test` — 1285 passing, no existing test needed adapting. `tsc`, `oxlint`, `oxfmt` clean.

New: `src/__tests__/relay-derived-stanza-node.test.ts`, 10 tests. Six on the predicate, including every rejection shape that must **not** be retried; four through `relayMessage` itself with a client that refuses the first attempt, asserting the retry drops only the named tag, keeps the caller's other nodes, sends once when nothing is refused, and never retries an unrelated rejection.

Reverting the retry drops two of them, both from the `relayMessage` group.

---
## Summary by cubic
Keeps caller-supplied `<biz>` working after upgrading to `@oxidezap/whatsapp-rust-bridge` 0.11.0. Old: duplicate derived-and-caller `<biz>` made sends fail or render nothing; new: we remove only the conflicting derived tag(s) the engine names and resend, so Baileys recipes send once and still work.

**Review notes**
- Bumps `@oxidezap/whatsapp-rust-bridge` to 0.11.0.
- Both chat and status relays go through a bounded retry that drops only the conflicting derived tag (`<biz>`, `<bot>`, etc.) named by the engine, repeating until no conflict; if the named tag is not present, the original error is rethrown.
- Unrelated errors are not retried; non-derived caller nodes are preserved.
- New helpers `sendDroppingDerivedNodes`, `derivedNodeConflictTag`, and `withoutDerivedNode` in `src/Compatibility/derived-stanza-nodes.ts`; wired in `src/Socket/messages.ts`.
- Tests added in `src/__tests__/relay-derived-stanza-node.test.ts`; no API or migration changes.

<sup>Written for commit fb0fbe92be5ae65b5739b323150e81c003ec34cf. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved message relay reliability when an engine-generated node conflicts with supplied message data.
  * Automatically retries affected relays after removing only the conflicting node.
  * Preserves unrelated nodes and continues to surface unrelated errors normally.

* **Tests**
  * Added coverage for conflict detection, selective node removal, retry behavior, and error handling.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@hapi/boom": "^9.1.4",
-        "@oxidezap/whatsapp-rust-bridge": "0.10.0",
+        "@oxidezap/whatsapp-rust-bridge": "0.11.0",
         "long": "^5.3.2",
         "pino": "^10.3.1",
         "protobufjs": "^7.6.5"
@@ -951,9 +951,9 @@
       }
     },
     "node_modules/@oxidezap/whatsapp-rust-bridge": {
-      "version": "0.10.0",
-      "resolved": "https://registry.npmjs.org/@oxidezap/whatsapp-rust-bridge/-/whatsapp-rust-bridge-0.10.0.tgz",
-      "integrity": "sha512-zsAR3hBw8EtnKq1pMfpXH2Q6b21BmTAv+LgQcAPazzga5BlcB0LQiQ7QeMLKNjoIgftlEmPNWMG5ELX12uWTlg==",
+      "version": "0.11.0",
+      "resolved": "https://registry.npmjs.org/@oxidezap/whatsapp-rust-bridge/-/whatsapp-rust-bridge-0.11.0.tgz",
+      "integrity": "sha512-k00OYftRzbHabPnCjZOadLZVuZcvnmoCJIh0Zt+4uvHqBFtf6LCIB8urhHSbEaR0uGbUDuWNFN0LGRFtWM2zwA==",
       "license": "MIT"
     },
     "node_modules/@oxlint/binding-android-arm-eabi": {
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
   },
   "dependencies": {
     "@hapi/boom": "^9.1.4",
-    "@oxidezap/whatsapp-rust-bridge": "0.10.0",
+    "@oxidezap/whatsapp-rust-bridge": "0.11.0",
     "long": "^5.3.2",
     "pino": "^10.3.1",
     "protobufjs": "^7.6.5"
--- a/src/Compatibility/derived-stanza-nodes.ts
+++ b/src/Compatibility/derived-stanza-nodes.ts
@@ -1,0 +1,49 @@
+import type { BinaryNode } from '../Types/index.ts'
+
+/**
+ * Upstream Baileys derives no stanza children of its own, so a caller that
+ * wants the `<biz>` on an interactive message has to attach it — which is what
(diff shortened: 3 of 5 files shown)


jlucaso1:
All three addressed.

**The unrelated-rejection test was asserting nothing** — good catch, and it was the worst of the three. It counted attempts on a client from a discarded `relayRefusing` instance, so the zero it checked was zero however the code behaved; a duplicated send would have passed. It now counts on the client under test and asserts exactly one attempt, which is what distinguishes propagating from resending.

**Multiple derived tags**: right, and it is the same incompatibility one step further in. A DM with an interactive payload derives both `<biz>` and `<bot>`, and the engine names one per attempt, so a caller supplying both had the first dropped and failed on the second. The retry is a loop now, bounded by the caller's own node list — each pass removes at least one node, so it cannot spin — and it gives up rather than resending if the refusal names a tag the caller never sent. Both cases have tests, and reverting the loop to a single retry drops one of them.

**Status relays**: also right. That branch returned before the retry existed. Both sends go through one shared path now, with a test covering the status one.

1288 passing.

PR #71, opened 2026-08-18
fix(messages): honour the caller's messageId on an edit

Salientekill:
## Problem

`options.messageId` overrides the outer stanza id on a plain send. On an **edit** it was silently dropped:

```ts
const newMsgId = await client.editMessageBytes(jid, protoMsg.key.id, editBytes)
```

The option never reached the bridge, so the edit always went out under the anchor message's own id. Observed from a caller passing `{ messageId }` on an edit:

```
pedimos 3EB0D274408CAD24EDCBDD, a lib usou 3EB0EB740CF66D956F65F3
```

The second id is the anchor's own — the request was dropped.

## Why it matters

The underlying Rust client supports this: `Client::edit_message_with_options` + `EditOptions::with_stanza_id`, documented there as letting callers control the outer stanza id, for example to collide it with an existing message so clients re-render that slot.

That collision is the only known way to take **payment requests** and **group status** messages off screen: the server ignores a plain revoke on those types, so overwriting the slot is the fallback. From Rust it worked; from JS it was unreachable.

Worth being precise about the ceiling, because it is easy to oversell: the core documents the collision as best-effort — the server may dedupe against the outer id, so the visible outcome is not guaranteed. And it does not make those types revocable. The official client gates "delete for everyone" on an allow-list of message types (`canSenderRevokeMsg` → `getIsUserCreatedType`) that `payment` is not in; the restriction is by type, not by id.

## Change

Forward `options.messageId` to `editMessageBytes` as the stanza id. It is the edit-path sibling of #72, which did the plain and status paths and left this one warning that it could not be honoured yet. That warning is gone.

**No cast.** Bridge 0.15.0 (already on `main` via #74) types the stanza id as a real fourth parameter, so the call is direct. An earlier revision reached for `as unknown as`, and it was worse than it looked: against an older bridge the wasm binding still takes three arguments and a fourth is dropped in silence, turning a visible "cannot honour this" into an edit that goes out under the wrong id. With the dependency bumped, the direct call typechecks and an older bridge fails the build instead of failing quietly.

**Deliberately not `resolveMessageId`.** The plain send path resolves an id for every send. Doing that here would be wrong: the engine treats *any* supplied stanza id as borrowed —

```rust
let borrowed_message_id = request_id.is_some();
```

— and binds no id-keyed state to it, so every edit would lose its retry-cache entry and outbound message secret, including edits nobody asked to pin. Absent stays absent, and `||` rather than `??` keeps an empty id meaning "unspecified" the way the rest of this API does.

## Compatibility

Omitting `messageId` keeps the previous behaviour byte for byte: no stanza id is passed, the engine assigns one, and the returned key carries it. No existing caller changes shape. Revokes are untouched and still warn, because the core has no options variant for a revoke to expose.

## Tests

The three cases move onto the harness that drives the real `sendMessage`. They previously exercised a copy of the forwarding declared inside the test file, so they passed whatever the production code did; reverting the change now fails them:

```
✖ forwards the caller id as the stanza id, leaving the edit target alone
ℹ pass 10  fail 1
```

Covered: an explicit id reaches the bridge as the fourth argument while the third still names the message being edited; no id means no fourth argument at all, which is what keeps an unasked edit from borrowing one; an empty id counts as absent.

## Validation

```
npm run format:check
npm run lint          # tsc + oxlint
npm test              # 1339 pass, 1 skip, 0 fail
npm run build
```


Salientekill:
Marked as draft: CI fails with `Expected 3 arguments, but got 4` because this depends on the bridge exposing the parameter — that is the companion PR **oxidezap/whatsapp-rust-bridge#69**, where `cargo check` passes.

Ready to un-draft as soon as the bridge change ships.

Salientekill:
Addressed: an empty `messageId` is now treated as unspecified (`||` instead of `??`), matching the falsy fallback used by `generateWAMessageFromContent` and `planMessageRelay`. Forwarding `''` would have pinned the stanza to an invalid id.

Salientekill:
Addressed what is actionable, and flagging what is not.

**Done:** `src/__tests__/edit-stanza-id.test.ts` covers the three shapes the review asked for — omitted `messageId` keeps the 3-argument call, an explicit id is forwarded as the stanza id, and an empty id counts as absent (matching the falsy fallbacks used by `generateWAMessageFromContent` and `planMessageRelay`).

**Blocked:** pinning a bridge version that declares the 4th argument. #70 landed it, but it is not published — `@oxidezap/whatsapp-rust-bridge` is still `latest = 0.14.0`, which types `editMessageBytes` with three parameters. There is no version to pin yet.

That is why the call is written to work against both: the 4th argument is used only when a caller passes `messageId`, through a narrowly typed alias rather than `any`. Against 0.14.0 the emitted call is byte-for-byte the previous one, which is why build, lint and tests are green here.

Happy to switch to a plain 4-argument call and pin the dependency in a follow-up once a bridge release with #70 is out.

jlucaso1:
@codex review

PR #82, opened 2026-08-21
perf(core): optimize jid normalization, content type detection and the CB attr walk

jlucaso1:
## Summary

Three hot paths on the message pipeline were allocating for work they never needed the allocations for. This removes the allocations and, where the shape allows it, the work.

- `areJidsSameUser` runs **3.0x-3.8x** faster and allocates **nothing** (was 44-84 B per call).
- `jidNormalizedUser` runs **3.4x-4.3x** faster; an already-normalized JID now allocates **nothing** at all (was 292 B), and the suffix-stripping path drops from 322 B to 21 B.
- `getContentType` runs **1.28x** faster and allocates **6 B** instead of 58 B per call.
- `emitCBEvents` drops the `Object.entries` pair array it built and discarded for every stanza the socket sees.

No public API changed and no observable behaviour changed. Every helper touched is covered by the differential fuzz suite against upstream Baileys, and the awkward edge cases are now pinned as unit tests too.

A fourth change, coalescing packed receipt batches into one `message-receipt.update` emission, was implemented, measured, and then **reverted**. See *Receipt batching, and why it is not here* below.

### `src/WABinary/jid-utils.ts`

`areJidsSameUser` and `jidNormalizedUser` both ran the full `jidDecode`, which slices the JID in two, splits the user half on `':'` and again on `'_'`, and boxes the pieces in a result object. Neither caller wanted the decoded parts, only the index the user component ends at: the first `':'` device suffix, `'_'` agent suffix, or the `'@'`.

- New `userEndBefore` finds that boundary with a bounded `charCodeAt` scan.
- `areJidsSameUser` compares the two user components in place. Zero arrays, zero objects.
- `jidNormalizedUser` no longer calls `jidDecode`/`jidEncode` at all. It slices at most once, and gains a fast path: a JID with no suffix to strip whose server is anything but `c.us` re-encodes to itself, so it is returned as-is with no new string built. That is the shape the pipeline mostly sees.

`String.prototype.split` with a one-character separator was the dominant cost here, at roughly 180 ns per call in this workload, and `jidDecode` made two of them per JID.

Behaviour is preserved down to the awkward corners, all verified against upstream `baileys@7.0.0-rc13` before and after: two JIDs that both fail to decode still compare equal (upstream compares two `undefined` users), a JID carrying more than one `'@'` keeps its extra one, an empty user still normalizes to a bare `@server`, and a separator with no server part after it still means "no user at all".

### `src/Utils/messages.ts`

`getContentType` built an `Object.keys` array and walked it with `Array.find`, allocating both the key array and the predicate closure on every inspected message, then ran a substring scan over every key it looked at.

- The scan is now an indexed loop with an early return, so the closure is gone.
- A `switch` settles the ten content keys that carry almost every message on a live socket before the substring scan is reached. Every one of those names already satisfies the upstream predicate, so this only shortcuts *how* the answer is reached, never *which* key wins.

The result stays positional (the first own key naming content), which `content-type-resolution.test.ts` pins in both directions and the fuzz suite compares against upstream.

One deliberate deviation from the original plan: the fallback stays `Object.keys` rather than `for..in`. A decoded `WAProto.Message` inherits a default for *every* field of the schema, so `for..in` enumerates ~100 prototype keys per call and would need a `hasOwn` filter to stay positional. Measured on decoded messages that is **746 ns/op versus 50 ns/op** for the code it would have replaced, a 15x regression on the exact input that dominates the real path. `Object.keys` yields only the handful of fields actually set, so its one small allocation is by far the cheaper of the two. A test pins that the inherited defaults stay invisible.

### `src/Socket/events.ts`

`emitCBEvents` drops the `Object.entries` pair array for a direct `for..in` over the attrs. Same walk, without the array it threw away for every stanza the socket sees.

The walk is guarded with `Object.hasOwn`. `Object.entries` was own-only by construction and `BinaryNode` is a public type, so an attrs map reaching the dispatcher with an enumerable on its prototype chain would otherwise fire `CB:tag,key:value` and `CB:tag,key` for an attr the stanza never carried. Attrs hold a handful of keys and the loop body emits two or three events per key, so the guard costs nothing measurable against what it protects. Pinned with a test that builds an attrs map via `Object.create({ inherited: 'ghost' })` and asserts no CB event mentions it.

## Receipt batching, and why it is not here

Coalescing a packed receipt batch into a single `message-receipt.update` emission was built and measured: 16x fewer EventEmitter round trips for a 16-receipt batch, 64x for a 64-receipt one, worth 3-10% of the batch handler's wall clock and 4-9% of its garbage.

It was reverted, because it traded a correctness property for that throughput.

`EventEmitter.emit` stops at the first listener that throws. With one emission per receipt, a consumer handler that blows up costs exactly the receipt it choked on and the loop carries on. Merged into one emission it costs the entire batch. Measured directly, on a three-receipt batch whose listener throws on the second:

| | second listener receives |
|---|---|
| one emission per receipt | `["A", "C"]` |
| one emission per batch | `[]` |

The exception stays contained either way and never escapes into the bridge, so this is not a crash. It is silent loss: receipts drive read and delivery state, and dropping up to sixty-four of them because one handler threw is a worse outcome than the few percent the merge bought on a handler already dominated by `decodeReceiptWireBatch` crossing into WASM.

Per-receipt isolation and a single emission are mutually exclusive, so the isolation stays. Per-listener isolation (the pattern `event-buffer.ts` already uses for `connection.update`) would not have rescued it either: one listener throwing part-way through a merged array still loses the remainder of the batch.

The property is now pinned by a test, so the emission count reads as a deliberate contract rather than an implementation detail left to tune later. Credit to the Codex review on this PR for catching it.

## Benchmarks &amp; Cost

Synthetic benchmark, Node v22.22.2, median of 5 passes of 5M iterations. Allocation is measured in a separate short GC-free window so `heapUsed` is a faithful counter. The benchmark script was temporary and is not part of this diff.

| Benchmark | Baseline | Patch | Speedup | Bytes/op |
|---|---|---|---|---|
| `areJidsSameUser` (normalized JIDs) | 683.6 ns/op &mdash; 1.46 M ops/s | 225.5 ns/op &mdash; 4.43 M ops/s | **3.03x** | 44 &rarr; **0** |
| `areJidsSameUser` (device/agent suffixes) | 809.5 ns/op &mdash; 1.24 M ops/s | 213.8 ns/op &mdash; 4.68 M ops/s | **3.79x** | 84 &rarr; **0** |
| `jidNormalizedUser` (normalized JIDs) | 356.4 ns/op &mdash; 2.81 M ops/s | 82.3 ns/op &mdash; 12.16 M ops/s | **4.33x** | 292 &rarr; **0** |
| `jidNormalizedUser` (device/agent suffixes) | 439.3 ns/op &mdash; 2.28 M ops/s | 122.2 ns/op &mdash; 8.18 M ops/s | **3.59x** | 322 &rarr; 21 (-93%) |
| `getContentType` (12 mixed contents) | 60.6 ns/op &mdash; 16.50 M ops/s | 47.5 ns/op &mdash; 21.03 M ops/s | **1.28x** | 58 &rarr; 6 (-90%) |

### Where the JID cost went

Isolating the pieces of `jidDecode` on the same machine: the two `String.prototype.split` calls cost **367 ns/op** between them, while the result object allocation costs 7 ns and the enum lookup 5 ns. Removing the splits is essentially the whole win.

## Validation

Run on this branch, all green:

- `npm run lint` (`tsc` + `oxlint`) &mdash; clean
- `npm run format:check` &mdash; clean
- `npm test` &mdash; **1372 passing**, 1 skipped, 0 failing (1367 before; 5 tests added)
- `npm run build` &mdash; clean
- `npm run compat:check-waproto` &mdash; clean
- `npm run compat:audit:proto --strict` &mdash; exit 0, findings unchanged from `main`
- `npm run compat:audit:wire --strict` &mdash; exit 0, 311 cases, 1970 preserved, 0 dropped, 0 altered; the 2 `pollResultSnapshotMessageV3` divergences are pre-existing and unrelated
- Differential fuzz vs upstream `baileys@7.0.0-rc13` (`pure-differential.fuzz.test.ts`, `bridge-events.fuzz.test.ts`) &mdash; 100 passing. `jidDecode`, `jidNormalizedUser`, `areJidsSameUser` and `getContentType` are all fuzz targets there.
- 45 hand-picked JID edge cases diffed directly against upstream before and after: all match.

### Tests added

- `public-helpers-compatibility.test.ts` &mdash; two cases pinning `areJidsSameUser` and `jidNormalizedUser` across device suffixes, agent suffixes, missing server parts on one or both sides, empty users, and multi-`@` JIDs.
- `content-type-resolution.test.ts` &mdash; pins that the schema defaults a decoded `WAProto.Message` inherits stay invisible to `getContentType`.
- `regressions.test.ts` &mdash; pins that an attrs map inheriting an enumerable fires no CB events for it, and that a throwing `message-receipt.update` listener does not discard the rest of a packed batch.

Both new negative tests were verified to fail against the implementation they guard against, not just to pass against the current one.

### Note on the branch name

Pushed to `claude/perf-jid-contenttype-receipts-r7ghbg` rather than `perf/jid-normalization-content-type-and-receipt-coalescing`: this session is pinned to that branch and cannot push elsewhere. Happy to have it renamed on merge.

### DIFF
--- a/src/Socket/events.ts
+++ b/src/Socket/events.ts
@@ -25,7 +25,7 @@ import type {
 	WhatsAppEvent,
 	WhatsAppEventCallbacks
 } from '@oxidezap/whatsapp-rust-bridge'
-import type { CanonicalEvent, CanonicalMessage, CanonicalReceipt } from '../Bridge/index.ts'
+import type { CanonicalEvent, CanonicalMessage } from '../Bridge/index.ts'
 import { adaptBridgeEvent, adaptBridgeMessageWire } from '../Bridge/index.ts'
 import { decodeHistorySyncWireBatch } from '../Bridge/history-sync-wire.ts'
 import { bridgeGroupMetadataToBaileys } from '../Compatibility/group-metadata.ts'
@@ -34,7 +34,6 @@ import type {
 	BaileysEventMap,
 	BinaryNode,
 	ConnectionState,
-	MessageUserReceiptUpdate,
 	WACallEvent,
 	WACallUpdateType,
 	WAMessage,
@@ -375,48 +374,6 @@ const describeTempBan = (code: number | undefined): string => {
 const emitConnectionUpdate = (ctx: SocketContext, update: Partial<ConnectionState>) =>
 	ctx.ev.emit('connection.update', update as Partial<ConnectionState>)
 
-/**
- * Append one `MessageUserReceiptUpdate` per message id (upstream emits per-id),
- * picking the timestamp slot from the receipt type so consumers branching on
- * `receipt.readTimestamp` / `receipt.playedTimestamp` see the right field
- * populated. Takes the sink so a packed batch can fill one array for the whole
- * batch instead of one array (and one emission) per receipt.
- */
-const collectReceiptUpdates = (evt: CanonicalReceipt, into: MessageUserReceiptUpdate[]) => {
-	const participant = evt.isGroup ? evt.senderJid : undefined
-	const receipt: { receiptTimestamp?: number; readTimestamp?: number; playedTimestamp?: number } = {}
-	if (evt.receiptType === 'read' || evt.receiptType === 'read-self') {
-		receipt.readTimestamp = evt.timestamp
-	} else if (evt.receiptType === 'played' || evt.receiptType === 'played-self') {
-		receipt.playedTimestamp = evt.timestamp
-	} else {
-		receipt.receiptTimestamp = evt.timestamp
-	}
-
-	for (const id of evt.messageIds) {
-		into.push({
-			key: { remoteJid: evt.chatJid, id, fromMe: evt.isFromMe, participant },
-			receipt
-		})
-	}
-
-	return into
-}
-
-/**
- * Release a collected receipt batch on the event channel, under the same
- * containment `dispatchCanonicalEvent` gave the per-receipt emission it
- * replaces: a throwing consumer must not escape back into the bridge. Kept at
- * module scope so a batch does not allocate a closure to flush itself.
- */
-const emitReceiptUpdates = (ctx: SocketContext, updates: MessageUserReceiptUpdate[]) => {
-	try {
-		ctx.ev.emit('message-receipt.update', updates)
-	} catch (err) {
-		ctx.reportUnexpectedError(err, "dispatching a 'receipt' event")
-	}
-}
-
 const DISPATCHERS: DispatcherMap = {
 	// ── Connection lifecycle ──
 	connected: (_, { ctx }) => emitConnectionUpdate(ctx, { connection: 'open' }),
@@ -648,7 +605,26 @@ const DISPATCHERS: DispatcherMap = {
 	},
 
 	receipt: (evt, { ctx }) => {
-		ctx.ev.emit('message-receipt.update', collectReceiptUpdates(evt, []))
+		// Fan out one MessageUserReceiptUpdate per id (upstream emits
+		// per-id) and pick the timestamp slot from the receipt type so
+		// consumers branching on `receipt.readTimestamp` /
+		// `receipt.playedTimestamp` see the right field populated.
+		const participant = evt.isGroup ? evt.senderJid : undefined
+		const receipt: { receiptTimestamp?: number; readTimestamp?: number; playedTimestamp?: number } = {}
+		if (evt.receiptType === 'read' || evt.receiptType === 'read-self') {
+			receipt.readTimestamp = evt.timestamp
+		} else if (evt.receiptType === 'played' || evt.receiptType === 'played-self') {
+			receipt.playedTimestamp = evt.timestamp
+		} else {
+			receipt.receiptTimestamp = evt.timestamp
+		}
+		ctx.ev.emit(
+			'message-receipt.update',
+			evt.messageIds.map(id => ({
+				key: { remoteJid: evt.chatJid, id, fromMe: evt.isFromMe, participant },
+				receipt
+			}))
+		)
 	},
 
 	undecryptableMessage: (evt, { ctx }) => {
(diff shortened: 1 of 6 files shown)


jlucaso1:
Addressed the merge-risk finding on the `emitCBEvents` walk in c9b9f43.

The point is correct and worth acting on rather than accepting: the `Object.entries` the loop replaced was own-only by construction, `for..in` is not, and `BinaryNode` is a public type whose `attrs` carries no guarantee of a bare prototype. An attrs map reaching the dispatcher with an inherited enumerable would have fired `CB:tag,key:value` and `CB:tag,key` for an attr the stanza never carried, which is exactly the kind of behaviour drift this PR is supposed to have none of.

The walk is now guarded with `Object.hasOwn`. The cost is nil in context: attrs carry a handful of keys and the loop body emits two or three events per key, each far more expensive than the check, so the allocation saved by dropping the pair array is kept intact.

Worth noting this is the same hazard the PR already reasons about in `getContentType`, where it is the reason that lookup deliberately stays on `Object.keys` instead of `for..in`. A decoded `WAProto.Message` inherits a default for every field of the schema, so there the prototype chain is not a hypothetical.

Pinned with a regression test that constructs an attrs map via `Object.create({ inherited: 'ghost' })` and asserts no CB event mentions it. Verified it fails without the guard and passes with it.

`npm run lint`, `npm run format:check` and `npm test` are green on the new head (1371 passing, 1 skipped).

---

jlucaso1:
Both review findings are resolved. Summary of where this landed:

**1. `emitCBEvents` own-attrs** (CodeRabbit inline + Codex P2) — fixed in c9b9f43. The walk is guarded with `Object.hasOwn`, restoring the own-only enumeration `Object.entries` gave for free while keeping the dropped pair array. Pinned with a test that builds an attrs map via `Object.create({ inherited: 'ghost' })`.

**2. Receipt-level dispatch isolation** (Codex P2) — confirmed and **the coalescing is reverted** in c3ccde4.

The finding was correct, and worth more than a caveat. I reproduced it directly on a three-receipt batch whose listener throws on the second receipt:

| | second listener receives |
|---|---|
| one emission per receipt | `["A", "C"]` |
| one emission per batch | `[]` |

`EventEmitter.emit` stops at the first listener that throws, so merging the batch turns "lose the one receipt that handler choked on" into "lose the whole batch". The exception stays contained and never reaches the bridge either way, so it is not a crash, it is silent loss — and receipts drive read and delivery state, so dropping up to sixty-four of them because one handler threw is worse than the 3-10% the merge bought on a handler already dominated by `decodeReceiptWireBatch` crossing into WASM.

Worth noting the two properties are mutually exclusive: per-receipt isolation requires per-receipt emission. Per-listener isolation, the pattern `event-buffer.ts` already uses for `connection.update`, would not have rescued it either, since one listener throwing part-way through a merged array still loses the remainder.

`onReceiptBatch` is therefore back to replaying the batch one receipt at a time, and there is now a test pinning it so the emission count reads as a deliberate contract rather than something left to tune later. Both new negative tests were checked to actually fail against the implementation they guard against, rather than merely passing against the current one.

The PR title and description have been updated: receipt batching is out of scope, and what remains is the JID work (3.0-4.3x, allocation eliminated), `getContentType` (1.28x, -90% bytes), and the `emitCBEvents` attr walk.

`npm run lint`, `npm run format:check` and `npm test` are green on c3ccde4 (1372 passing, 1 skipped, 5 tests added).

---

PR #127, opened 2026-09-09
perf: drop protobufjs runtime dependency with structural facade shim

jlucaso1:
Removes the 3.9 MB install-time cost of shipping protobufjs to every consumer even though no shipped JS imports it at runtime.

What changes:
- scripts/compatibility/waproto-facade.ts now strips the upstream 'import * as $protobuf from protobufjs' line and emits a full-fidelity structural mirror of the pinned protobufjs 8.8.0 Writer/Reader instance surface plus an identical IConversionOptions. Constructors and statics are intentionally omitted: no facade signature references either class constructor or static side, which also keeps Buffer (used only by Reader.create) and its @types/node requirement out of the consumer contract.
- src/WAProto/index.d.ts regenerated via compat:sync-waproto; lib/WAProto/runtime.d.ts carries the same bytes through the existing copy-build step.
- protobufjs demoted to devDependencies (still used by the facade generator and differential tests); the now-unneeded overrides pin is dropped. Baileys dev-only resolves its own nested protobufjs 7.6.6.
- New scripts/compatibility/__tests__/consumer-no-protobufjs.test.ts proves a consumer program typechecks with protobufjs unresolvable (isolated temp dir, negative control included).

Compatibility evidence (Baileys-first):
- scripts/compatibility/type-contracts.ts ProtoNamespaceExact (line 54) passes unweakened; the shim is mutually assignable with the real classes.
- npm test: 1527 pass, 0 fail, 1 skipped. npm run lint, format:check, build, check-pack, and check-registry-deps all pass. npm ls --omit=dev shows no protobufjs.
- Pre-existing failures left untouched: type-contracts.ts:138-139 fail identically on base; compat:layers needs a sibling core checkout absent here.

Follow-up stays upstream-only: @bufbuild/protobuf reaches consumers transitively through the bridge dist (bundled, type-only) and can only be demoted there.

---
## Summary by cubic
Removes `protobufjs` from the runtime dependencies so consumers no longer install it, replacing the import in the generated WAProto declaration with a structural type shim. No runtime behavior changes: no shipped JS imports `protobufjs`.

- The facade generator strips the upstream `protobufjs` import and emits a structural mirror of the `Writer`/`Reader`/`IConversionOptions` surface matching `protobufjs` 7.6.6 (the version Baileys consumers resolve), omitting constructors and statics so `Buffer` and `@types/node` stay out of consumer typechecking.
- `protobufjs` moves to `devDependencies`, the `overrides` pin is dropped, and a new test proves a consumer program typechecks with `protobufjs` unresolvable, including cross-compat with the nested 7.6.6 types.
- The Exact-facade compatibility assert still passes, so the shim stays mutually assignable with the real `protobufjs` classes.

<sup>Written for commit cfa72adda05eee521d1e4abbff25c32a5655511e. Summary will update on new commits.</sup>

## Summary by CodeRabbit

- **Compatibility**
  - The published WAProto interface no longer requires `protobufjs` to be installed at runtime.
  - Consumers can continue using message creation, encoding, decoding, conversion, and type URL APIs without a direct runtime dependency on `protobufjs`.
  - Type compatibility with applications using `protobufjs` 7.x remains supported.

- **Tests**
  - Added compatibility coverage to verify package usage when `protobufjs` is unavailable at runtime.


jlucaso1:
@codex review

jlucaso1:
Fixed in cfa72ad: the shim now mirrors the protobufjs 7.6.6 instance surface that Baileys rc14 resolves (verified: nested 7.6.6 has head/tail/states:object|null, no pos/buf/raw/packed helpers/finishInto/tag, skipType without fieldNumber), not the top-level 8.8.0.

One subtlety the mirror reproduces exactly: 7.6.6 declares its own minimal { low, high, unsigned } Long for the Reader/Writer methods while message fields keep the full long import, so the shim nests a shadowing minimal Long inside $protobuf for the same split. Real 7.6.6 readers/writers now flow both directions across encode/decode in the consumer-typecheck fixture (which now compiles against the actual nested 7.6.6 declaration as ground truth, plus the protobufjs-absent negative control), and type-contracts.ts:54 Exact passes unweakened. Full suite: 1527 pass, 0 fail.

jlucaso1:
CI Tests failure is a timing flake, not a regression from this PR: the LRU test asserts a TTL-20ms entry survives two 12ms sleeps, and a loaded runner stretches the timer past expiry (undefined !== 1). Evidence: the test file plus everything it imports (src/Compatibility/internal/*) is byte-identical between base and this branch (only a comment differs in proto-runtime.ts, which the test never imports; the rest of this diff is types/manifest), and the file passes 30/30 runs locally. Re-ran the failed CI job (run 34406387063, job 102650304434).

PR #131, opened 2026-09-09
feat(events): carry the envelope's type on an undecryptable stub

Salientekill:
## Summary

The stub emitted for a message that failed to decrypt says it is a CIPHERTEXT placeholder and, when the server named one, why it is unavailable. It does not say what the message *was* — even though the bridge hands that over on every one of them: `MessageInfo.type` is the envelope's `type` attribute (`"text" | "media" | "pay" | "poll" | "reaction" | "event"`), read from the stanza in the clear.

That attribute is the only thing about an unreadable message a consumer can still trust:

- it describes the **stanza**, not the ciphertext, so a decryption failure does not take it away;
- the server stamps it on the **sender's own** outgoing stanza, so it cannot be pointed at somebody else — unlike a `contextInfo.participant` in a quote, which names a third party and needs corroboration.

## Why it came up

Group moderation. A payment stanza whose sender deliberately withholds the sender key from one device arrives there as a placeholder that device cannot classify at all — the admins never see the content, and the bot holding the connection has nothing to act on. The signals left are a third party quoting the message, or a human reporting it: both are after the fact, and both need anti-forgery checks precisely because they point at someone else.

`type="pay"` on the envelope needs none of that. The same client on the Rust side reads `StanzaMessageType::Pay` off `Event::UndecryptableMessage` and acts on it directly; a JS consumer cannot, only because this handler drops the field.

## Changes

- `Bridge/types.ts` — `stanzaType` on the undecryptable event.
- `Bridge/schema.ts` — the adapter reads `info.type`, which it already had in hand.
- `Socket/events.ts` — it rides in `messageStubParameters` beside `unavailableType`, which is where the stub already carries envelope facts. Both are filtered for presence, so a stanza with neither still emits `[]` and nothing changes for a consumer that reads neither.

## Tests

Two beside the existing `undecryptable_message` cases: the type alongside an `unavailable_type`, and alone when the stanza carries none.

```
npx tsc -P tsconfig.build.json --noEmit    clean
npm test                                   1534 passed, 0 failed
```

## Not verified

No live account here, so nothing proves end to end that a real withheld-sender-key payment arrives with `type="pay"` — that claim rests on the bridge's own doc comment for `MessageInfo::type` and on the Rust client reading it from the same event.

---
## Summary by cubic
Carries the envelope's `type` on undecryptable message stubs so consumers can still classify a message whose content failed to decrypt.

Previously the stub only said it was a CIPHERTEXT placeholder and, when supplied, why it was unavailable. Now the stub carries `stanzaType` — the server-stamped envelope `type` (`"text" | "media" | "pay" | "poll" | "reaction" | "event"`) as its own field. It survives decryption failure and can't be attributed to a third party.

- Adds `stanzaType` to `CanonicalUndecryptableMessage` and reads it from `info.type` in the schema adapter.
- Emits it as a named field on the stub; `messageStubParameters` is untouched, so existing consumers see no shape change.
- Adds tests for the type with and without an `unavailable_type`.
- Documents the divergence from upstream in the README's Gotchas.

**Not verified**
- End-to-end delivery of a real withheld-sender-key payment stanza was not tested; the type's presence relies on the bridge's `MessageInfo.type` contract.

<sup>Written for commit 56c745e7b01b7d0d286b3f366a133165710414cd. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Undecryptable messages now retain their original message type, such as text, media, payments, polls, reactions, or events.
  * Placeholder ciphertext messages now include this type alongside any unavailable-message details, improving message classification when decryption fails.

* **Tests**
  * Added coverage for undecryptable messages both with and without unavailable-message type information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,6 +237,18 @@ preserved. No QR re-scan, no logged-out events.
 
 A few behaviors that differ from upstream — almost always to your advantage:
 
+- **An undecryptable message tells you what it was.** The CIPHERTEXT stub
+  emitted on `messages.upsert` for a message that failed to decrypt carries
+  `stanzaType`: the envelope's `type` attribute as the server stamped it on the
+  sending stanza — `"text"`, `"media"`, `"pay"`, `"poll"`, `"reaction"`,
+  `"event"`. It describes the stanza rather than the ciphertext, so it survives
+  the failure, and it is stamped on the sender's own outgoing stanza, so it
+  cannot be pointed at anyone else. Upstream has no equivalent: there the
+  placeholder is unclassifiable. `messageStubParameters` is untouched and still
+  holds only the unavailability reason, when the server named one — the two are
+  independently optional, which is exactly what a positional array cannot
+  express.
+
 - **Auto-reconnect is built in, but `close` still means `close`.** The Rust
   engine retries transient drops on a fibonacci backoff and reports them as
   `connection: 'connecting'`, so the canonical upstream handler never fires
--- a/src/Bridge/schema.ts
+++ b/src/Bridge/schema.ts
@@ -255,6 +255,7 @@ const ADAPTERS = {
 			remoteJidAlt: resolveRemoteJidAlt(senderAlt, recipientAlt, isGroup, isFromMe),
 			isUnavailable: asBoolOr(data?.is_unavailable, false),
 			unavailableType: asString(data?.unavailable_type),
+			stanzaType: asString(info.type),
 			decryptFailMode: asString(data?.decrypt_fail_mode),
 			raw: data
 		}
--- a/src/Bridge/types.ts
+++ b/src/Bridge/types.ts
@@ -571,6 +571,13 @@ export interface CanonicalUndecryptableMessage {
 	isUnavailable: boolean
 	/** Bridge `unavailable_type`: "view_once" | "unknown". */
 	unavailableType?: string
+	/**
+	 * The envelope's `type` attribute, as the server stamped it: "text",
+	 * "media", "pay", "poll", "reaction", "event"… It describes the stanza,
+	 * not the ciphertext, so it survives a decryption failure — the one thing
+	 * a consumer can still learn about a message it cannot read.
+	 */
+	stanzaType?: string
 	/** Bridge `decrypt_fail_mode`: "show" | "hide". */
 	decryptFailMode?: string
 	/** Original raw payload, preserved for debug logging. */
--- a/src/Socket/events.ts
+++ b/src/Socket/events.ts
@@ -619,17 +619,18 @@ const DISPATCHERS: DispatcherMap = {
 			messageTimestamp: evt.timestamp,
 			pushName: evt.pushName,
 			messageStubType: WAProto.WebMessageInfo.StubType.CIPHERTEXT,
-			// The envelope's `type` rides along with `unavailableType`. It is
-			// the only thing that survives a decryption failure to say what the
-			// message was — the server stamps it on the sender's outgoing
-			// stanza, in the clear — and dropping it here left a consumer with
-			// a placeholder it could not classify at all.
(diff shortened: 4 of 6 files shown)


jlucaso1 (changes requested):
Thanks for this, it is a well put together PR. I checked the mechanism, the tests, and the compat audits. The approach is sound. I am requesting changes on a few points before merge.

Required: `npm run format:check` fails on `src/Socket/events.ts:627-629`. CI enforces it, so this blocks the merge. The formatter wants the filter on one line:

```ts
messageStubParameters: [evt.unavailableType, evt.stanzaType].filter((value): value is string => !!value)
```

No behavior change, just formatting.

Two things the automated reviews did not catch. First, `src/Bridge/types.ts:572` documents `unavailableType` as `"view_once" | "unknown"`, but the bridge type also allows `"hosted"` and `"bot"`. Since you are touching this area, please bring the comment in line with the real type. Same file, the stanza type list at 574-579 uses an ellipsis where the bridge contract has exactly seven known values plus a string fallback. Name them or point at `StanzaMessageType`.

Second, a note on the trust level of the new field. The envelope type is stamped by the sender, so a moderation consumer should read `stanzaType` as a self-declared hint, not proof of what the message is. It cannot frame someone else, but it can mislabel its own message. One sentence in the code comment would stop a future reader from building an enforcement decision on it alone.

Optional and not blocking: the fuzz generator never emits `info.type`, so fuzz will not exercise this path. Adding `type` to the `infoBlock` draw in `src/__fuzz__/generators/bridge-event.ts` would close that gap.

The bot reviews already covered the parameter ordering discussion, so I will not repeat it here beyond saying I agree the array growth is acceptable for a `feat`.

Salientekill:
Both P1s are the same defect and they are right: filtering makes the position mean different things depending on what the stanza carried. With no `unavailable_type`, `"pay"` lands at index 0 — the slot that until now only ever held an unavailability reason — and since both fields are free strings, a consumer reading `[0]` cannot tell which one it got. I introduced an ambiguity where there was a single meaning, and the third comment follows from it: that is a consumer-visible contract change with nothing documenting it.

I would rather you pick the shape than guess and spend another round on it:

1. **Fixed slots** — `[unavailableType ?? '', stanzaType ?? '']`, emitted whenever either exists. Positions stay stable and readable by index, at the cost of an empty string standing in for absence, and of `[]` becoming `['', 'pay']` for a stanza that only has the type.
2. **Its own field** — leave `messageStubParameters` exactly as it is today and hang the envelope type off the stub instead (beside `key.participantAlt` / `key.remoteJidAlt`, which are already non-`WebMessageInfo` facts the stub carries). Nothing existing shifts; the new fact is unambiguous by name rather than by position.

I lean towards 2: `messageStubParameters` is positional by nature, and a second independently-optional value is exactly what positional arrays are bad at. It also keeps the change additive — no existing reader observes anything different.

Whichever you prefer, I will add the README note for it in the same push. Sorry for the noise on the first cut.

Salientekill:
Went with shape 2, pushed.

`messageStubParameters` is byte-for-byte what it was before this PR: the unavailability reason when the server named one, `[]` otherwise. The envelope type is now `stanzaType` on the stub, beside `key.participantAlt` and `key.remoteJidAlt` — facts the stub already carries that `WebMessageInfo` does not model. Two independently optional values are what a positional array cannot express; a name can, and it makes the change purely additive: no existing reader observes anything different, which also settles the undocumented-contract-change comment at its root.

Typed on `WAMessage` with the values it can take and the note that it is set only on the CIPHERTEXT stub, and documented in the README's Gotchas, where the other consumer-visible divergences from upstream live.

```
npx tsc -P tsconfig.build.json --noEmit    clean
npm test                                   1534 passed, 0 failed
```

The two tests now assert both halves: the type is on the field and the array is left alone, with and without an `unavailable_type`.

jlucaso1 (approved):

PR #136, opened 2026-09-10
feat(calls): consume the bridge encoded-audio domain with JS pumps

jlucaso1:
Real audio calls in baileyrs, on top of the bridge calls preview. Dial and answer
encoded-audio calls, push packets, collect decoded ones, read stats, hang up.

Which bridge ref this uses

I checked all three places. Bridge main has no calls domain at all, so there
was nothing to consume there. The pkg.pr.new build for PR 115 first served an
older push with signaling only, then refreshed to the PR head once its CI went
green. package.json points at the PR URL and the lockfile pins the exact build
I verified against, 0.0.0-preview-2071901, which carries both the signaling
slice and the encoded-audio slice. Never a release, as agreed.

What this adds

A new src/Socket/calls.ts with the audio half of the socket. dialCall and
acceptCall open a call with an mlow or opus promise, pushCallAudio queues one
packet and answers true or false, endCall reports how much of the peer was
told, and setCallMuted, getCallMediaStats, getActiveCalls and
setRelayTransportProvider round it out. I did not reimplement the relay
transport or any SDP text. The provider passes straight through to the bridge,
so the bridge stays the only place that knows WebRTC.

Two pieces around those calls. A media router takes the bridge onCallAudio
and onCallEvent callbacks, fans decoded frames out to per-call sinks, and
emits a call.media event for lifecycle steps including ended, which also stops
that call's pumps. And a pump runner pulls packets from a source into the
push, counts shed audio instead of erroring on it, and stops on a spent
source, an abort, or the call ending.

For tests without a microphone there are two sources and no new dependencies.
The silence source sends the single 0x90 byte the core accepts as an encoded
payload under both format promises, and the file source chunks a fixture into
fixed packets. Neither decodes anything. There is deliberately no ffmpeg in
this package.

Backpressure works the way the bridge contract says. A full engine queue sheds
newest-first and the push answers false, which the pump counts and reports
through onShed. The bridge exposes no watermark readout, its own docs name the
false return as the pacing signal instead, so this layer paces off the shed
count plus the audioSinkDropped and inboundPipeDropped stats counters. Frames
cross without an extra copy on this side. The bridge already paid the one copy
each way and the router hands the frame view straight to the sink.

Without a bridge that has the client-calls-audio domain, every method throws
501 naming the capability instead of failing on a missing function. The bridge
types for the new ops are written down structurally in calls.ts so this still
compiles against the release bridge, and a runtime probe picks the real client
up when the preview is installed.

Evidence

I ran the loop for real against the local Bartender mock with two paired
clients on the preview build. Alice dialed Bob's LID and got a call id back,
Bob's call event carried the offer with the same id, three silence pushes
queued true, the stats came back with all 21 counters numeric, and endCall
reported peer-notified. A one-packet silence pump through startCallAudioPump
resolved pushed 1 and shed 0. That run is now a committed e2e test and it
passed three times in a row here. It needs the mock, so fork CI skips it like
the other e2e files.

Around it, the offline suite covers what needs no peer. call-audio-pump has 24
tests for the sources, the pump totals, shed accounting, sink routing and the
method validation, all passing. call-audio-surface drives the real preview
WASM offline and checks every op is present with its sync or async shape and
that unknown ids and bad arguments come back as typed invalid-argument errors.
I also extended the two coverage ledgers this change trips, the closed-domain
list and the fuzz boundary table, with dialCall and acceptCall audioFormat
cases.

Compatibility is unchanged. The missing, wire, proto and lifecycle audits
return the exact same finding sets on base and head, so zero new divergences.
lint, format check, build and check-pack are green. The full unit suite is
1569 of 1571.

Two things I did not do

Decoded audio coming back down onCallAudio over a live relay is still
unproven. The mock answers no relay, so the media plane stays dormant and no
packet can come back that way here. The inbound half is covered by the loopback
tests, and a live relay handshake needs an account-backed rig that belongs to
the next slice. The native voip-cli target was not reachable for the same
reason, it needs real accounts.

The one failing unit test needs its own note. bridge-free-safety fails under
the preview with a wasm memory access out of bounds when free lands with a
fetchBlocklist in flight. I reproduced it on the untouched base tree with only
the preview installed, and the same tree passes 4 of 4 on the release bridge,
so this is a property of the preview build, not of this change. The fix belongs
on the bridge side and I left the assertion alone.

compat:layers did not run. It looks for the core and bridge checkouts at fixed
sibling paths that do not exist in this worktree layout, so I report it blocked
rather than passed.

---
## Summary by cubic
Adds encoded-audio voice calls on top of the bridge calls preview. Previously only call signaling flowed; now `dialCall`, `acceptCall`, `pushCallAudio`, `endCall`, `setCallMuted`, `getCallMediaStats`, and `getActiveCalls` drive the bridge's `client-calls-audio` domain, with per-call audio sinks, a pump runner, and shed accounting.

**Dependencies**
- Swaps `@oxidezap/whatsapp-rust-bridge` from release `0.21.4` to the pkg.pr.new preview for PR 115, pinned to `0.0.0-preview-2071901` in the lockfile.
- This is never a release, and a release build without the audio domain makes every method throw `501` naming `client-calls-audio`.

**Behavior notes**
- Backpressure is loss-tolerant: a `false` push means the packet was shed, counted and reported via `onShed`, never an error.
- The media router hands frames to sinks without an extra copy and stops calls' pumps on `ended` or socket teardown.
- Decoded audio over a live relay is unproven — the mock answers no relay, so that direction stays covered by loopback tests.
- `bridge-free-safety` fails under the preview build with a wasm memory access out of bounds, reproduced on the untouched base — a preview-build property, not this change.
- `compat:layers` did not run because it needs core and bridge checkouts at fixed sibling paths this worktree lacks.

<sup>Written for commit fa363e95279a8ee8a6ecec880d1e6ebb60fa0356. Summary will update on new commits.</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,6 +314,22 @@ A few behaviors that differ from upstream — almost always to your advantage:
   other media peers (`sharp`, `jimp`, `audio-decode`, `link-preview-js`).
   Without it, audio messages still send — they just carry no `seconds`
   value. Install `music-metadata` if you want durations computed.
+- **Voice calls ride the bridge calls preview, not the release bridge.**
+  `sock.dialCall(peerJid)` / `sock.acceptCall(callId)` open an encoded-audio
+  call (`'mlow'` by default, `'opus'` for the in-profile escape);
+  `sock.pushCallAudio(callId, bytes)` queues one packet and resolves `false`
+  when the engine shed it under backpressure — that is the normal
+  loss-tolerant answer, not an error. Decoded packets arrive through
+  per-call `sock.onCallAudio(callId, sink)` sinks, lifecycle steps
+  (`relay-allocated`, …, `ended`) on the `call.media` event, and
+  `sock.endCall` / `setCallMuted` / `getCallMediaStats` / `getActiveCalls`
+  round out the surface. There is no watermark readout: the bridge names the
+  `false` return as the pacing signal, with the shed counters in the stats
+  for the rest. `startCallAudioPump(callId, source)` wires a packet source
+  to the push; the built-in silence (`0x90` MLOW SID) and file-chunk sources
+  need no microphone, codec, or ffmpeg and are meant for tests. Calls need a
+  bridge with the `client-calls-audio` domain — a preview build, not a
+  release — and without one every method above throws `501` naming it.
 - **Your key store also holds bridge state, so "empty" is not "unpaired".**
   See [Bridge state in your key store](#bridge-state-in-your-key-store) — this
   one can break a boot path, so it has its own section.
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@hapi/boom": "^10.0.1",
-        "@oxidezap/whatsapp-rust-bridge": "0.21.4",
+        "@oxidezap/whatsapp-rust-bridge": "https://pkg.pr.new/@oxidezap/whatsapp-rust-bridge@115",
         "long": "^5.3.2",
         "pino": "^10.3.1"
       },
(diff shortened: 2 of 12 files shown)


jlucaso1:
Closing in favor of https://github.com/oxidezap/baileyrs/pull/134, which now carries both scopes. The audio commit from this branch was rebased onto the 134 branch as f28e451, the combined tree re-ran green (lint, format, unit, build, pack, identical audit sets, e2e call loop green locally with the full suite left to CI), and the 134 body now covers the signaling and audio halves with API examples. From here the whole VoIP line in this package stays on 134 alone.

PR #143, opened 2026-09-13
feat(history): wire history sync admission policy

jlucaso1:
## Summary

- Update `@oxidezap/whatsapp-rust-bridge` to release `0.22.0`.
- Translate Baileys `shouldSyncHistoryMessage` into the bridge's generic `historySyncAdmission` policy.
- Reproduce the upstream decision order exactly: the callback runs first, then the `PROCESSABLE_HISTORY_TYPES` gate rejects unknown or absent sync types.
- Reconstruct a decoded `HistorySyncNotification` carrying only pre-download metadata (`syncType`, `chunkOrder`, `progress`, `fileLength` as `Long`, `peerDataRequestSessionId`); absent fields stay non-own, matching protobufjs shape.
- Resolve an explicit `undefined` policy to the default; warn once at construction when the policy disables every processable type (upstream parity, including the synthetic `{ syncType }` probe calls).
- Mark the policy as read configuration instead of reporting it as unsupported.

## Scope

This PR resolves admission policy only, not full history-sync parity: `shouldSyncHistoryMessage` sees a partial notification view (no media/key fields), the policy gates received notifications but not the initial-sync state machine, and requesting full history (`syncFullHistory`) stays engine-owned.

## Validation

- `npm run build`
- `npm run lint`
- `npm run format:check`
- `npm test`: 1563 pass, 0 fail, 1 skipped
- Focused suites: adapter mapping/shape/gate/probe/fallback plus socket plumbing (ninth bridge argument, `undefined` fallback, all-disabled warning).

## Summary by CodeRabbit

* **New Features**
  * Added configurable history synchronization admission using the existing `shouldSyncHistoryMessage` policy.
  * The policy now evaluates available history-sync metadata before content is downloaded or parsed.
  * Unsupported history-sync types are rejected automatically after policy evaluation.
  * `shouldSyncHistoryMessage` is now recognized as a supported socket configuration option.
* **Documentation**
  * Expanded configuration guidance to describe available metadata, unsupported fields, and acknowledgment behavior.
* **Bug Fixes**
  * Improved handling of history-sync progress, ordering, file length, and session information.


jlucaso1:
All nine review items are addressed in `59e47b7`:

1. **Processable gate** — the adapter now calls `shouldSyncHistoryMessage` first, then ANDs the result with `PROCESSABLE_HISTORY_TYPES.includes(...)`, matching `chats.js:931-933` order. Unknown/absent `syncType` is rejected even when the callback accepts.
2. **Partial view** — documented as a contract decision, not silent: `SocketConfig.shouldSyncHistoryMessage` docs, README gotcha, and PR scope section all state the callback sees only pre-download metadata with media/key fields absent. Nothing is synthesized.
3. **`undefined` fallback** — `resolveHistorySyncPolicy` coalesces to the default; the socket resolves against the raw config before the shallow merge can clobber it.
4. **Protobuf shape** — notifications are built with `HistorySyncNotification.create()` from only-present fields: `instanceof` holds, absent fields are non-own (`Object.keys` empty), reads match the protobufjs `null` prototype default. Tests assert all three.
5. **All-disabled warning** — `isHistorySyncFullyDisabled` probes the policy once per processable type with synthetic `{ syncType }` (same call pattern as `socket.js:33`), and the socket warns with the upstream text when everything is rejected.
6. **README** — `shouldSyncHistoryMessage` removed from the unsupported list; new gotcha documents the partial view, permanent ack, engine-owned `syncFullHistory`, and the initial-sync lifecycle difference.
7. **Contract tests** — new `history-sync-policy-plumbing.test.ts` builds the real socket offline: ninth bridge argument accepted, `undefined` policy initializes, all-disabled warns, default stays quiet. Adapter suite covers gate order, shape/prototype, probe, and fallback.
8. **Lifecycle note** — documented in README and PR scope: unlike upstream, the policy does not gate the initial-sync state machine; admission covers received notifications only.
9. **`syncFullHistory`** — stays in `UNSUPPORTED_CONFIG_KEYS` and engine-owned, called out explicitly in the PR scope section.

Validation: `npm run build`, `npm run lint`, `npm run format:check` clean; `npm test` 1563 pass / 0 fail / 1 skipped.

jlucaso1:
Reply to the `isHistorySyncFullyDisabled` finding (`src/Compatibility/history-sync-admission.ts:68`):

Partially accepted and fixed in `a9e8035`. The probe itself stays: it is verbatim upstream parity (`socket.js:33` calls `shouldSyncHistoryMessage({ syncType })` once per `PROCESSABLE_HISTORY_TYPES` at construction), explicitly requested in human review, so stateful policies observing those calls matches upstream exactly — removing the calls would be the divergence, not keeping them.

What was valid in the finding: a probe call that throws used to propagate out of `makeWASocket`. The probe is diagnostic-only, and this repo guarantees a socket never fails to build over a diagnostic, so each probe call is now guarded — a throw counts as potentially enabling (no warning) instead of breaking construction. This is a deliberate, documented deviation from upstream, which lets it throw. Added a regression test for the throwing policy.

PR #146, opened 2026-09-17
test(fuzz): compare deep values exactly instead of summarising them

jlucaso1:
## Summary

Related to #56.

Below the old 12-level recursion cutoff, `normalise` replaced subtrees with summaries of their key names, so two differently wrapped values with different leaves compared equal — the exact question the deep differential inputs are built to ask. The summary also lost the schema path, so `isTextField` could receive the wrong prefix, and `omitsKeysOnly` shared the lossy view, letting a deep value change read as a field omission.

Replaces the lossy summaries with an exact, stack-safe iterative traversal in `compare.ts`, preserving the existing API and policies: scalar coercion, `-0` handling, presence, bytes/Date/Map/Set/Long treatment and Map ordering are unchanged, and schema paths stay intact at any depth. Internal markers are now branded out-of-band, so literal `{ __bytes__: ... }`-shaped data can no longer be confused with one. Shared children compare by value (a DAG is not a cycle); a real cycle is rejected explicitly; traversal past a work budget throws instead of returning a partial or equal result. `omitsKeysOnly` now walks the same single comparison, returning true only for a genuine key omission with no changed shared value.

## Scope

Fuzz-harness comparator only; no divergence predicate, codec or production library code changed. Cyclic inputs were previously rendered as a `<cycle>` summary and now raise `TypeError` — the suite never fed cycles to the comparator, and an equality result for them was never trustworthy.

## Validation

- New unit suite `compare.test.ts` (46 tests) failed before the fix: false equality at depths 13/14/400, stack overflows on long chains, marker collisions, cycle handling, budget behaviour and deep Map ordering. All pass after it.
- Probe: 14 `{ next: ... }` wrappers over `{ a: 1, b: 2 }` vs `{ a: 9, b: 2 }` now compare unequal in both default and strict modes.
- `npm test`: 1611 passed, 0 failed, 1 existing GC-dependent skip.
- `npm run lint`
- `npm run format:check`
- `npm run build`
- `node scripts/check-pack.ts`: 297 files, no duplicates.
- `git diff --check`

No nightly deep campaign or live E2E was run; E2E needs the private Bartender mock.

---
## Summary by cubic
Fixes the fuzz harness comparator to traverse deep values exactly instead of summarising subtrees past a depth cutoff, so differently wrapped values with different leaves now compare unequal and schema paths survive any depth.

- Replaces lossy summaries with an exact, stack-safe iterative traversal.
- Preserves existing comparison policies: scalar coercion, `-0`, presence, bytes/Date/Map/Set/Long handling, and Map ordering.
- Internal markers are branded out-of-band so literal objects like `{ __bytes__: ... }` can't be confused with generated markers.
- Shared children compare by value; real cycles now throw `TypeError` instead of being rendered as a `<cycle>` summary.
- Traversal past the work budget throws instead of returning a partial or equal result.
- `omitsKeysOnly` now walks the same single comparison and returns true only for a genuine key omission with no changed shared value.
- No production code changed; only the fuzz-harness comparator and its test suite.

<sup>Written for commit c027755d2eeeee407f9ef2965aa82beebfc6341a. Summary will update on new commits.</sup>

## Summary by CodeRabbit

- **Tests**
  - Added comprehensive comparison-harness coverage for cyclic data, traversal limits, deeply nested structures, and complex keys.
  - Added checks distinguishing generated markers from literal objects.
  - Added strict-option tests covering scalar coercion, value presence, numeric types, special values, containers, shared references, and Map/Set ordering.


jlucaso1:
@coderabbitai review

PR #149, opened 2026-09-17
fix(proto): translate the bridge's renamed wire fields in the facade and send path

jlucaso1:
## Summary

Two of the sixteen wire gaps the strict proto audit tracks were name mismatches,
not missing fields. The bridge regenerated its schema against a newer WhatsApp
snapshot, so these kept their number and their wire type and changed spelling:

| Type                             | Baileys                  | Bridge                           | Field |
| -------------------------------- | ------------------------ | -------------------------------- | ----- |
| `Message.ExtendedTextMessage`    | `faviconMMSMetadata`     | `faviconMmsMetadata`             | 33    |
| `Message.MessageHistoryMetadata` | `oldestMessageTimestamp` | `oldestMessageTimestampInWindow` | 2     |

Measured, not assumed: encoding each with the bridge spelling and decoding with
the pinned `baileys` codec (and the reverse) produces identical bytes and lands
on the same field, so the wire already agreed and only the property name differed.
What that cost was silent: a Baileys-spelled write was dropped by the bridge with
no error, and a decode never set the public property, so code written against the
upstream types read `undefined`.

Both spellings are now accepted on the way in, with an own public field winning
over the bridge name, and decoded values, `toJSON()` and `toObject()` report the
public spelling, nested occurrences included. Inputs are not mutated.

## The other half of the round trip

The facade is only where messages are *read*. The send path encodes through the
neutral codec (`encodeProtoCompat` → `encodeProto`), and that codec drops a key it
does not know without throwing — so translating the decode side alone would have
made things worse: a message decoded through the facade and then relayed
(`relayMessage(jid, msg.message)`) or re-sent lost both fields, where before it kept
them under the bridge's spelling. A public-spelled field written by the caller never
reached the wire either. This is what an independent review of the first revision of
this branch found, and it is fixed here rather than shipped.

`encodeProtoCompat` now projects the message before encoding. The projection is
copy-on-write and returns the same reference when nothing is aliased, so a message
without these fields takes the path it always did. Repair-on-throw still runs only
after a failure, and now from the caller's message rather than the projected one,
because the repair walks the schema this library publishes — the projected object
carries the codec's names, which it would not find, and an empty-string int64 would
have propagated instead of being absorbed. The repair also accepts either spelling
for an aliased field now, since the projection is what hands the codec the bridge's
name; before that, `faviconMmsMetadata` with a refused nested value threw where the
public spelling encoded.

Three harnesses compare a send against a plain encode, and all three now account for
the documented renames instead of reporting the translation as a defect:

- `scripts/compatibility/wire-fidelity-core.ts` projects the reference message,
  without which three of its cases reported a false 'altered'.
- `src/__fuzz__/wire-fidelity.fuzz.test.ts` projects its `wire:fidelity` reference
  too — that one is directional, so an unprojected reference left a projection that
  drops the field invisible from both sides.
- The same file's `wire:upstream-readable` target folds `RENAMED_PROTO_FIELDS` out of
  both decoded views of the same bytes, which is what the codec differentials already
  do.

A fourth check covers what the projected reference cannot: the wire-fidelity audit now
also asserts the value the caller planted reaches the decoded bytes, an expectation
that comes from the input rather than from either encode.

The neutral bridge keeps renaming both fields, so `RENAMED_PROTO_FIELDS` and the
`NOT_ENCODED_FIELDS` entry in `src/__fuzz__/harness/divergence.ts` still describe
the raw codec accurately: the fuzzers compare `encodeProto`/`decodeProto` from the
bridge directly, which is a different axis from this facade audit.

This is the same class of fix as #147, which covered the two `deviceID` spellings,
and it closes on the write path the defect #147 left open.

## Validation

- `node --test src/__tests__/proto-field-aliases.test.ts` — 15 tests, 15 pass. Each
  aliased field is checked against the pinned `baileys` codec per input (encode
  bytes, `decode().toJSON()`, `toObject`, own-property presence, non-mutation), plus
  two fixed wire fixtures (`10 07`, `8a 02 00`) so a rename cannot hide inside a
  shared round trip.
- `node --test src/__tests__/encode-proto-compat.test.ts` — 29 tests, 29 pass,
  including the relay case (bytes the facade decodes, re-encoded through the send
  path against hand-written wire bytes) and the refused value under either spelling.
- `node --test scripts/compatibility/__tests__/wire-fidelity.test.ts` — 8 tests, 8
  pass, including the planted-value check.
- `npm test` — 1635 pass, 1 skipped (E2E), 0 fail.
- `npm run lint`, `npm run format:check`, `npm run build`, `node scripts/check-pack.ts`
  (297 files) — pass.
- `npm run compat:audit:proto -- --strict` — wire fields **2410/2424 (99.42%)**,
  `unexpectedGaps` and `resolvedKnownGaps` both empty, with the two fields pinned by
  name in the compatibility test so a later change cannot reopen one silently. The
  entries left `KNOWN_WIRE_GAPS` because the audit reports them matched and refuses
  to pass otherwise (2408 → 2410 matched, 16 → 14 gaps).
- Cost on the send path, measured here: the projection itself is 60–720 ns per
  message, from a plain `conversation` to a nested aliased field. An A/B against the
  previous implementation of `encodeProtoCompat` on messages without aliased fields
  stayed inside this machine's noise (±10% across fifteen paired rounds); the aliased
  case costs about 1.1 µs more because it now encodes a field the old path dropped.

## Not verified

- No WhatsApp Web capture exercises these two fields from this client; the reference
  is the pinned `baileys` codec, and both field numbers were cross-checked against
  the schema the bridge consumes.
- Live E2E was not run locally (private Bartender mock); it is the skipped test above
  and runs in CI.
- The `wire:upstream-readable` fold rebuilds every object it walks, so a byte field
  reaches the comparison as a numeric-keyed plain object on both sides. Byte identity
  is therefore no longer something that comparison can fail on; field numbers,
  presence and values still are. It is applied to both sides for exactly that reason —
  folding one side would compare that object against a `Uint8Array`.

## Notes

The `README.md` section from #147 did not land in the merge, so the documentation for
the renamed fields — now covering all four and both directions — is restored here.

## Scope

The fourteen remaining gaps are not reachable from this layer: six `mediaKeyDomain`
fields and `SyncActionValue.businessBroadcastAssociationAction` are absent from the
schema the bridge consumes (`businessBroadcastAssociationAction` is field 65, which
the bridge codec does not know at all), `BotAvatarMetadata` and
`BotMetadata.avatarMetadata` have no bridge codec, and
`Message.pollResultSnapshotMessageV3` is a deliberate field-number difference where
the bridge matches WhatsApp Web (115) and upstream declares 114.

Related to #56

---
## Summary by cubic
Fixes the protobuf field aliases between Baileys and the bridge so renamed fields (`faviconMMSMetadata`/`faviconMmsMetadata`, `oldestMessageTimestamp`/`oldestMessageTimestampInWindow`) round-trip correctly: writes are no longer dropped and reads set the public spelling.

- Encoding and construction (`encode`, `fromPartial`, `fromObject`) accept either spelling, with an own public field winning, and every read-out reports the Baileys spelling.
- `encodeProtoCompat` projects aliases before encoding so relayed or re-sent messages keep all four renamed fields, and the repair path accepts either spelling too.
- The wire-fidelity harnesses project or fold the documented renames to match the send path, and the audit now also checks the delivered value.
- The strict proto audit now matches 2410 of 2424 wire fields (was 2408), with the renamed fields pinned by name in compatibility tests and documented in `README.md`.

<sup>Written for commit c015c09cd0cf3b9c3c2b588129c2b4cac292022c. Summary will update on new commits.</sup>

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed compatibility for several protobuf fields whose public and bridge spellings differ.
  - Messages now correctly preserve these fields during encoding, decoding, partial construction, relaying, and JSON/object conversion, including nested values.
  - Explicitly supplied public properties take precedence when both spellings are present, while input objects remain unchanged.

- **Documentation**
  - Added guidance explaining supported field-name translations and compatibility behavior.


jlucaso1:
@coderabbitai review

jlucaso1:
**Review finding on `fromObject`/`toObject`: verified and applied** in `c015c09`.

The gap was real, though not a parity bug: `encode` and `fromPartial` accepted a bridge-spelled aliased field while the two object converters dropped it silently — the same failure this PR removes on the write path. Checked against the pinned `baileys` first: upstream drops it too (`{}` for a bridge-spelled input), so accepting it is additive rather than a change to any object upstream already accepts.

What the commit does:

- `fromObject` reads the bridge key when the public property is not an own property.
- `toObject` reports the value under the public name, and its "the instance carries this field" test accepts either spelling now.
- `hasOwn` rather than a null check: an instance carries its defaults on the prototype, so an absent public field reads as `0` or `null` there and the bridge key would never be reached.
- An own public property still wins, including an explicit `null`/`undefined`, which omits the field.

Validation: `src/__tests__/proto-field-aliases.test.ts` 19/19 — each aliased field is checked with upstream's own `fromObject`/`toObject` as the oracle for the name and the conversion, plus a both-spellings case; `npm test` 1639 pass / 1 skipped / 0 fail; lint, format:check, build, check-pack and both compatibility audits green.

Not done: the suggested `coderabbit review --agent` run. I have no verified local setup for it, and the change is covered by the tests above.

PR #150, opened 2026-09-17
docs(compat): state which reference wins for wire and surface decisions

jlucaso1:
## Summary

Documentation only — no behaviour changes.

The compatibility boundary translates between two references that disagree, and the
rule for which one decides what was implicit. It is now written down, in the
directory that owns the translation:

- **The wire follows the client.** Field numbers, presence and encodability come
  from WhatsApp Web — extracted by [whatspec](https://github.com/oxidezap/whatspec),
  carried through `whatsapp-rust` into the bridge.
- **The public surface follows Baileys.** Property names, optionality and object
  shapes stay a drop-in for consumers even where the bridge spells a field
  differently; those translations are `FIELD_ALIASES` in `proto-runtime.ts`.

So a disagreement that is only a name is this layer's job, and a disagreement about
which field exists or at which number is the client's to settle.

## What the rule settles, and the evidence now recorded beside it

`KNOWN_WIRE_GAPS` in `scripts/compatibility/proto-runtime-audit.ts` lists gaps whose
reason lived only in earlier investigation. Each is a field the WhatsApp Web schema
no longer declares while the pinned `baileys` still does. Checked against the client
itself, not only against baileys:

| Source | `mediaKeyDomain` ×6 | `businessBroadcastAssociationAction` | `BotAvatarMetadata` |
| --- | --- | --- | --- |
| whatspec extraction, WhatsApp Web 2.3000.1047483476 | only on `MediaDomainInfo` | type declared, no field references it | absent |
| bridge schema manifest | same | same | absent |

`SyncActionValue` numbers run 60–64 then 66, leaving a hole exactly where baileys
declares 65, and no bot-avatar type is declared anywhere. Nothing in whatspec,
`whatsapp-rust` or the bridge can add them back, and output that did would put a
field on the wire the client does not read.

One substitution was following Baileys rather than the client and did not say so:
`encodeProtoCompat` turns an empty string in a 64-bit field into `0`, where the
client never writes it and the bridge refuses it deliberately. It stays — a message
that used to reach the server should not start throwing — and it is now documented
as Baileys compatibility rather than client compliance, beside the surrogate case
that does follow the client.

## Validation

Documentation only: the two `.ts` files touched carry comments, and `README.md`
prose. No runtime behaviour changed, so no runtime test is evidence for this diff.

- `npm run lint`, `npm run format:check` — pass.
- `npm run compat:audit:proto -- --strict` — unchanged, wire fields 2410/2424
  (99.42%), no unexpected or resolved gaps, which is the check the edited comments
  sit in.
- `node --test src/__tests__/proto-runtime-compatibility.test.ts` — 7 tests, 7 pass
  (the suite that asserts those audit totals).

Related to #56

---
## Summary by cubic
Documents the previously implicit compatibility rule: the wire follows WhatsApp Web, while the public surface follows Baileys. This keeps client-only schema gaps upstream-owned and preserves Baileys-compatible names and shapes without changing runtime behavior.

- Records the client and bridge evidence for the six legacy `mediaKeyDomain` fields, `SyncActionValue.businessBroadcastAssociationAction`, and `BotAvatarMetadata`.
- Clarifies that `FIELD_ALIASES` handles naming differences, while field existence and numbers come from the client schema.
- Documents the intentional `encodeProtoCompat` empty-string substitution as Baileys compatibility, alongside client-aligned surrogate handling.
- Addresses #56; lint, formatting, the compatibility audit, and the related test suite pass.

<sup>Written for commit 0781b256027355752a1657aa145721be3b1a7df9. Summary will update on new commits.</sup>


jlucaso1:
Folded into #151, which now carries this commit alongside the prose trim. The rule and the evidence belong with the trimmed comments they explain, so they ship as one pull request.

